PR #4, opened 2017-12-07
Add a link to video_player

mit-mit:
Thanks for this great plugin! Just a small PR to add a link to `video_player`


mit-mit:
cc @brianegan

brianegan:
Good call, thanks for the contribution!

PR #24, opened 2018-06-07
Support latest version of video_player

cbenhagen:


brianegan:
Awesome, thanks for the update @cbenhagen. I'll publish a new version with this update

PR #27, opened 2018-06-11
A fix to allow using the controller between chewie players

zshema:
remove _onPlay listener from the controller when the widget is disposed and avoid disposing the controller that I'm not the owner of it - the owner is responsible to invoke dispose() (and potentially reuse it ).


brianegan:
Good call -- I was working on something similar to fix another bug, this is a good first step.

PR #28, opened 2018-06-14
Fix time formatting

cbenhagen:
Displayed as 02:010 before


brianegan:
Awesome, thanks for the Bugfix!

PR #29, opened 2018-06-16
Fix skipping

cbenhagen:
It now really skips 15 seconds as intended


brianegan:
Thanks for all these updates. I'll be back from a short trip tomorrow and will publish all of them as a new version!

cbenhagen:
You're welcome ;) Thanks for the plugin!

PR #32, opened 2018-06-19
 Start video at a certain position

cbenhagen:
Maybe the parameter should be renamed to `startAt`?.


brianegan:
Thanks! Yep, I'd recommend `startAt` or `startPosition` instead of `seekTo` -- I could see folks getting confused by that one.

cbenhagen:
@brianegan I renamed it.

brianegan:
Thanks again for the contributions :) I'll publish a new version with all your updates

Issue #31, opened 2018-06-19
Fade between placeholder and first video frame


cbenhagen:
Right now it goes from placeholder to black to first frame which looks pretty crappy. Maybe we can't do much about the first frame being black but fading would at least make the experience a lot smoother.

brianegan:
Gotcha -- I understand the idea, but might have to play with this one a bit more to make sure it "feels" right and doesn't interfere with video playback while the transition is going on (if the transition was 300ms, the user might not see some frames in the video).

cbenhagen:
I'd say 150ms might be enough. Maybe much more important is to get rid of that flashing black frame before the video starts. Could need to be fixed in the video_player plugin..

PR #34, opened 2018-06-21
WIP: Fadeout placeholder

cbenhagen:
Proposition to fix https://github.com/brianegan/chewie/issues/31

This solves the black frame at the beginning as well. One thing though: I can see a one pixel (or less?) line of the video loading in the background on the left and right side of the clip even when setting `StackFit.expand`. @brianegan do you have an idea why?


cbenhagen:
One way to fix the lines might be to just have the same order as before and do a fade-in of the video instead of a fade-out of the placeholder but i'd still love to know why it happens :P

brianegan:
Gonna close this for now as I think it's been open for a bit and might not apply to the current version :)

PR #43, opened 2018-07-26
Added CircularProgressIndicator on first launch

bostrot:
Slow loading videos will just show an empty progress bar which is irritating for most users. This adds a [CircularProgressIndicator](https://docs.flutter.io/flutter/material/CircularProgressIndicator-class.html) on first buffering.


brianegan:
Hey there -- thanks for the contribution! Hrm, good point! My only thought: I don't know if hard-coding a `CircularProgressIndicator` is the best solution to this problem (especially on iOS).

I wonder if we could instead take in a `Widget loading` property on the `Chewie` constructor, which would allow folks to specify the type of loading state they want?

If the user doesn't provide anything, fallback to `CircularProgressIndicator` on Android, or on iOS use the `CupertinoActivityIndicator`?

Thoughts?

PR #56, opened 2018-10-03
resolves video_player version & adds support for starting video on fullscreen 

miguelpruivo:
This PR will:
- Keep the `video_player` version below 0.6.4 which is fine for video rendering since there is a known issue with 0.6.5 that prevents frames to be shown unless the video enters in full screen and then disposes back to inline.
- Adds a property to choose whether the video should start on full screen or not when the playback starts.


brianegan:
Thanks for the contribution! I'll merge this in and bump the video_player version up to 0.7.0+ which I believe also fixes those problems.

brianegan:
Published as part of `0.8.0`. Thanks again!

PR #58, opened 2018-10-05
Always wrap player in AspectRatio to preserve aspect ratio in fullscreen

patrickbz:
Wrapping the player also when in fullscreen preserve the correct aspect ratio of the video. It seems to work well on Android and iOS. Thank you


brianegan:
Thanks!

brianegan:
Published as part of `0.8.0`. Thanks again!

Issue #64, opened 2018-10-22
Sleep in fullscreen


romatroskin:
Device sleep when playing video in fullscreen, there is https://pub.dartlang.org/packages/screen, and it's should be like 1 line to add :)

Sub6Resources:
We should probably have a parameter that allows us to choose whether or not we want the screen to be allowed to sleep.

cbenhagen:
This is fixed in #67

Issue #44, opened 2018-08-09
Hide or disable progress bar


marianoarga:
Hello, how can I hide the progress bar, since I'm streaming live video and dont need it.
I have commented "_buildPosition" and "_buildProgressBar" but It'st really hacky and mess with controls distribution.

Could you add that feature?
Thank you.

aqzwss:
+1

cbenhagen:
Future versions of Chewie will make it easier to customize the controls. I will close this in favor of #100.

PR #67, opened 2018-11-12
Added options for live and screen sleep

romatroskin:
In this PR I've added two options to hide time and progress bar which is called "isLive" and another one to allow keep device screen on while in full screen (allowedScreenSleep).
Issues: #64 #44


brianegan:
Looks great, thanks!

PR #75, opened 2018-12-12
Correct video link in code example

nstrelow:
Video link was moved. Corrected it.


nstrelow:
Is included in #79 , since it is so small.

PR #79, opened 2018-12-16
Dispose VideoPlayerController when Widget is disposed

nstrelow:
Prevents the video from still playing in the background, when ChewieWidget is disposed.

Example:
When I close the view containing the ChewieWidget, e.g. navigate somwhere, the audio can still be heard in the background, since the VideoPlayerController has not been disposed.

Also includes the small Readme link fix from #75

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ import 'package:chewie/chewie.dart';
 
 final playerWidget = new Chewie(
   new VideoPlayerController.network(
-    'https://flutter.github.io/assets-for-api-docs/videos/butterfly.mp4'
+    'https://flutter.github.io/assets-for-api-docs/assets/videos/butterfly.mp4'
   ),
   aspectRatio: 3 / 2,
   autoPlay: true,
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -111,8 +111,8 @@ class _ChewiePlayerState extends State<Chewie> {
   
   @override
   void dispose() {
-    super.dispose();
     _controller.dispose();
+    super.dispose();
   }
 
   Widget _buildFullScreenVideo(
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -108,6 +108,12 @@ class _ChewiePlayerState extends State<Chewie> {
     _controller = widget.controller;
     _initialize();
   }
+  
+  @override
+  void dispose() {
+    super.dispose();
+    _controller.dispose();
+  }
 
   Widget _buildFullScreenVideo(
       BuildContext context, Animation<double> animation) {


lodemo:
Hello,
thanks for the feature addition, i would need this for my use case as well.

I have tried the code snipped for the dispose() method and got an AlreadyDisposed Exception. 
I am using the player in a list view (SliverList) and the video muted correctly when scrolling away, but when scrolling to the view again the exception occurred.

nstrelow:
For me and I think for a lot of other users, this is critical and I hope it can be merged and released as soon as possible. Still, it's the 21th, so who cares for 2018, hope it @brianegan can address this in 2019. Happy Holidays 🎄

I-am-original:
@brianegan this issue may be not so critical, but it is cool to have controller disposing, however, @nstrelow  you can do disposing on your side, as you have reference to controller.

@brianegan I would like to add one more PR with aspectRatio but have no access rights to do that

nstrelow:
Yeah, I a currently doing that, but it feels unclean to have a State only to dispose the videoPlayer onDispose

brianegan:
Thanks for the PR -- back from vacation.

I definitely understand why you'd want this -- it's actually how Chewie originally worked! Unfortunately, it caused problems for many users. Why? Many folks do not want Chewie to dispose of the controller automatically, and need to re-use the controller.

Since it seems some folks want this and others don't, perhaps we could add an option like `autoDispose` set to `false` by default? If it's true, it will dispose of the controller.  If it's false, it will allow the user to manage the controller.

I-am-original:
@brianegan  as for me I would left it as it is. When you need something and now how to do - it is cool. If you need not - it is bad to have some useless code in Chewie. My opinion is like that )

nstrelow:
I can imagine the videoController is reused when building some kind of picture in picture mode, where the videoPlayer is removed from Chewie and added somehwere else?

The default would need to be false to not break existing code. Even though IMO the autoDispose = false is the special case from my understanding.

@brianegan As this is your package, it will be your call. Whether you add autoDispose (with default false) or not, a small entry in the Readme would be nice, explaining, that the player needs to be disposed to stop playing. And then instructions for either using the `autoDispose` option or a small example how to call this in a StatefulWidget (or maybe there is even a better way).
I will be happy to make a PR for this, if you don't find the time 😁

lodemo:
Im currently using Chewie in a dedicated StatefulWidget and dispose the controller in the widget's dispose method. This works fine for my use case in which the widget is placed inside a listview, and as long the video is watched "inline". 

However as soon as you open fullscreen mode the widget's dispose is called, thus the controller is disposed and the video playback in fullscreen won't work.

Not sure how to handle this, and i think this would be the same case with this pull request wouldn't it?

```
[ERROR:flutter/shell/common/shell.cc(184)] Dart Error: Unhandled exception:
E/flutter (30083): A VideoPlayerController was used after being disposed.
E/flutter (30083): Once you have called dispose() on a VideoPlayerController, it can no longer be used.
E/flutter (30083): #0      ChangeNotifier._debugAssertNotDisposed.<anonymous closure> (package:flutter/src/foundation/change_notifier.dart:108:9)
E/flutter (30083): #1      ChangeNotifier._debugAssertNotDisposed (package:flutter/src/foundation/change_notifier.dart:114:6)
E/flutter (30083): #2      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:202:12)
E/flutter (30083): #3      ValueNotifier.value= (package:flutter/src/foundation/change_notifier.dart:264:5)
E/flutter (30083): #4      VideoPlayerController.play (package:video_player/video_player.dart:293:5)
```

nstrelow:
@lodemo I actually have no crash, but as soon as I go into fullScreen and back out again, dispose is not called anymore and you can hear the sound of the video, even if you leave the view.

Not sure if calling dispose in chewie would fix this. I have to try it, but it could be possible.

lodemo:
It seems that the List which contains the widgets with Chewie are rebuild when entering fullscreen mode and again rebuild when exiting, so disposed is called multiple times.

I think it has to do with how the tree can be rebuild after push/pop see https://github.com/flutter/flutter/issues/18366 and https://github.com/flutter/flutter/issues/11655

Which makes me wonder how to correctly handle/dispose the controller in a StatefulWidget if the Widget could be rebuild anytime, while also not breaking fullscreen mode?


Maybe i should open a issue to discuss this further.

nstrelow:
@lodemo Defenitly make an issue for it. Maybe there is a way to make a fix in the plugin or a workaround can be found. Since more people should face the same issue

cbenhagen:
The current README mentions to dispose both controllers. I think there are situations where it would be a bit easier if Chewie would dispose the controllers. But I think more often you would not want it to do this for you. I am pro making it optional with a default set to `false`.

@nstrelow Please rebase and add an option for `autoDisposeVideoPlayerController`. You might also want to add `autoDisposeChewieController`.

nstrelow:
@cbenhagen Awesome, saw the new README entry. Also nice 0.9 release.
Out of curiosity, what are use cases for "not disposing automatically"?

I will build in the suggested changes, when I find the time 👍

cbenhagen:
You might want to have some sort of playlist where you have multiple VideoControllers which should keep their playback state. Or you if you implement your own full screen logic. Also picture in picture comes to mind. @brianegan might have other examples where users did not want chewie to automatically dispose. It might in fact be worth to think about it some more to decide which one should be the default.

brianegan:
Hey all, I'm gonna close this one out! I can understand why folks might want this, but it just caused to many problems in the past.

PR #99, opened 2019-01-16
ChewieController

### LINKED ISSUE #1: Make the code more manageable
Code needs a good refactor / clean up.

### LINKED ISSUE #23: Launch video in fullscreen through a custom button.
Hello,
I wanted to hide the video controls and alternatively launch the video in fullscreen when the user clicks on a specific button. I know that you can hide the controls through showControls attribute but I was wondering if there was anything to change the video to fullscreen. I do not think that the video_player plugin has anything of such functionality either although it does have a setVolume property which is missing in chewie. I was hoping if you knew anything about setting the video to fullscreen without relying on the video controls or could create such functionality.
Thank You.

### LINKED ISSUE #49: Fullscreen events
Hello, is there any way to access the events when the video comes out of full screen?
Because my application runs in full screen and leaves it when I leave a video in full screen.
I need my application to return to full screen.
Thanks!
Sorry for my english.

cbenhagen:
This is an attempt in fixing #1 by implementing a `ChewieController` which is internally accessed by an `InheritedWidget`.

TODO:
- [x] Check if playback hack is still needed and if yes, add it back in
- [x] `VideoPlayer` tries to access disposed `VideoController` after we changed and disposed it in the `ChewieController`
- [x] Add data to be rendered on all the Widgets, probably as `Listenables`, such as `ChangeNotifier` or `ValueNotifier`.
- [x] Update changelog
- [x] Document how to migrate from `Chewie` < 1.0.0

fixes #1 
fixes #23 
fixes #49


cbenhagen:
@brianegan I know it still has some rough edges but if you have time it would be awesome if you could have a quick look at the changes and tell me if this is the direction you'd like this to go.

lib/src/chewie_controller.dart
```diff
@@ -0,0 +1,182 @@
...
+  void play() {
+    value.videoPlayerController.play();
+  }
+
+  void pause() {
+    value.videoPlayerController.pause();
+  }
+
+  // TODO: Do we really need the ability to change the controller?
+  set videoPlayerController(VideoPlayerController controller) {
```
brianegan: I would say probably not -- if someone wants to show a different video, I think it makes sense to dispose of the ChewieController and create a new one.

Things like AspectRatio, etc might not be handled correctly for different videos

lib/src/chewie_controller.dart
```diff
@@ -0,0 +1,182 @@
...
+
+  @override
+  String toString() {
+    return '$runtimeType('
+        'isFullscreen: $isFullScreen, '
+        'videoPlayerController: $videoPlayerController, ';
+  }
+}
+
+class ChewieController extends ValueNotifier<ChewieValue> {
```
brianegan: Could `ChewieController` be a `ChangeNotifier` instead of a `ValueNotifier<ChewieValue>`, and move the `isFullScreen` and `videoPlayerController` as part of it?

That way folks could just do `chewieController.isFullscreen` or `chewController.addListener`, etc. 

That way you could just access those

brianegan:
Other than a couple small nitpicks, this is exactly the direction I was thinking of going. Really well done :D :D :D

lib/src/chewie_controller.dart
cbenhagen: It was modelled after the `VideoPlayerController` which extends `ValueNotifier<VideoPlayerValue>`. It has getters like the `controller.position` but other values are accessed like `controller.value.isPlaying` or `controller.value.volume`. But you are right it would be a nicer API if it can all be accessed as short as possible. I will look into using `ChangeNotifier` directly.

brianegan: Ah, you're right... If you have the ability, maybe take 5-10 minutes to quickly refactor it and see how it works out! 

If the current API you've got here seems fine, no worries at all.

lib/src/chewie_controller.dart
cbenhagen: Sounds good to me.

Up until now this was being done in `Chewie`: https://github.com/brianegan/chewie/blob/b135a466622daf080282b756cf6749844ed05632/lib/src/chewie_player.dart#L180-L185

Maybe we should adapt the example to show the proper disposing of both of the controllers. Or is there a better way we can help the user?

brianegan: I think we should generally encourage folks to dispose of both separately, because it seems like there are a lot of cases where folks do NOT want Chewie to do that automatically for them (and where it causes bugs).

What are your thoughts?

lib/src/chewie_controller.dart
cbenhagen: I could also add getters to the current `ChewieController`:
```dart
bool get isFullScreen => value.isFullScreen;
```
Will try to find out why some controllers out there are using `ValueNotifier` and others `ChangeNotifier`.

brianegan:
Two more things: As part of this PR, could you please add yourself to the list of authors in the `pubspec.yaml`? I'll also give you uploader rights for this repo so once this is ready to roll you can go ahead and publish it :)

cbenhagen:
Done :)

brianegan:
Cool, just added that email as an uploader as well. You should get a confirmation email to that address! 

Only thing I'd ask: Please update the CHANGELOG as well when ya push new versions to let folks know what's changed. Other than that, not too much to publishing new versions!

cbenhagen:
Sure! I will also not push bigger changes without discussing them first. And then we should always guide the users how to migrate. Thank you for that great package!

lib/src/chewie_controller.dart
cbenhagen: Added the two getters. So the API is nice and short now. If at some point we should see big advantages in using the `ChangeNotifier` directly we can switch without breaking stuff for users.

lib/src/chewie_controller.dart
cbenhagen: Let's leave it to the user for the moment.

lib/src/chewie_controller.dart
cbenhagen: Switched to using a `ChangeNotifier`. We can switch back to using ValueNotifier when we have a lot more values and it is more convenient to change and notify with one `value.copyWith(foo='bar')` instead of explicitly having to call notifyListeners(). Until then the code is just much cleaner without so much boiler plate.

cbenhagen:
@brianegan, you once mentioned to "add data to be rendered on all the Widgets, probably as `Listenables`, such as `ChangeNotifier` or `ValueNotifier`".  What do you think, should we add this to this release? If yes, what data did you have in mind?

brianegan:
Nope, this covers it -- the idea was to give something folks can listen to -- in this case, the `ChewieController` does the job, and Widgets can get the info they need from that controller (or the nested VideoController)!

cbenhagen:
@brianegan I got more plans for Chewie like supporting multiple versions of a clip as well as playlists. I guess they can both be added without breaking the API again by adding named constructors and internally switching to always having a list(of lists) of VideoControllers but setting one as the current `videoPlayerController`. So unless you see any breaking changes ahead, you could do a final review of this PR so we can publish this.

lib/src/chewie_controller.dart
```diff
@@ -0,0 +1,165 @@
+import 'package:chewie/src/chewie_progress_colors.dart';
+import 'package:flutter/material.dart';
+import 'package:video_player/video_player.dart';
+
+class ChewieController extends ChangeNotifier {
```
brianegan: Can we please add some DartDocs (e.g. `/// Triple slash coments`) to this, even if it's just a line or two?

pubspec.yaml
```diff
@@ -1,8 +1,10 @@
 name: chewie
 description: A video player for Flutter with Cupertino and Material play controls
-version: 0.8.0
+version: 1.0.0
```
brianegan: What do you think about going to 0.9.0 for now and 1.0.0 in a month or so -- that will allow us to get public usage & feedback before we declare this code stable?

lib/src/material_controls.dart
```diff
@@ -68,47 +57,41 @@ class _MaterialControlsState extends State<MaterialControls> {
...
     _showAfterExpandCollapseTimer?.cancel();
   }
 
   @override
-  void initState() {
+  void didChangeDependencies() {
+    chewieController = ChewieControllerProvider.of(context);
+    controller = chewieController.videoPlayerController;
+
+    _dispose();
```
brianegan: Unlike `initState`, `didChangeDependencies` can run multiple times throughout the lifecycle of the Widget. Therefore, we may want to add some logic around this `_dispose` / `initialize` code: Only run that dispose and init IF the `controller` has changed.

lib/src/chewie_controller.dart
```diff
@@ -0,0 +1,165 @@
...
+    Key key,
+    @required this.controller,
+    @required Widget child,
+  })  : assert(controller != null),
+        assert(child != null),
+        super(key: key, child: child);
+
+  final ChewieController controller;
+
+  static ChewieController of(BuildContext context) {
```
brianegan: Can we move this method to the `ChewieController` class? That will make it so you can access the ChewieController via `ChewieController.of` instead of `ChewieControllerProvider.of`. This matches the same style as other "Inherited" data, such as `Theme.of`.

We could even make this provider private! Such as `_ChewieControllerProvider` and use it only internally.

lib/src/chewie_controller.dart
cbenhagen: Sure, can you come up with a short and precise text for it? I tried but postponed it..

pubspec.yaml
cbenhagen: The intention was not to declare it stable, but to make it clear that there are incompatible API changes. If we choose to follow semver we have to update the major version number on such breaking changes.

brianegan: Ah, gotcha -- IMO, 1.0.0 releases have that feeling of "ok, we think this is stable now."

And It's totally cool if we change the API before 1.0 according to SemVer in a `0.9.0` release -- pre 1.0 "Anything may change at any time." That, and I've been using the MINOR version (x.MINOR.y) to denote breaking changes on this lib for a bit now :P

lib/src/chewie_controller.dart
brianegan:
```
/// The ChewieController is used to configure and drive the Chewie Player Widgets. It
/// provides methods to control playback, such as [pause] and [play], as well as methods
/// that control the visual appearance of the player, such as [enterFullScreen] or 
/// [exitFullScreen].
///
/// In addition, you can listen to the ChewieController for presentational changes, such as
/// entering and exiting full screen mode. To listen for changes to the playback, such as 
/// a change to the seek position of the player, please use the standard information provided 
/// by the `VideoPlayerController`.
```

How does that sound?

pubspec.yaml
cbenhagen: Ok. But starting from 1.0.0, if we break the API we move to 2.0.0?

lib/src/chewie_controller.dart
cbenhagen: Sounds good to me. Thanks!

lib/src/chewie_controller.dart
cbenhagen: `ChewieController.of` is indeed much nicer. I somehow missed that this could be done in this way.

pubspec.yaml
brianegan: Yep, exactly. So my thinking: put this at 0.9.0 for the time being. See if we get any more feature requests / bug reports that would require breaking change in the next month or so. If not, we can ship 1.0.0, and if we need to make breaking changes beyond that it'd require a major version bump.

lib/src/material_controls.dart
cbenhagen: Right, it was also reinitializing when the theme changed. fc1bc16 fixes that.

cbenhagen:
Just noticed that I need a gmail account to upload. Please add my personal mail as uploader: c.ben.hagen@gmail.com

brianegan:
Oh dang -- just added that email address!

PR #110, opened 2019-02-07
Update to support video_player >=0.10.0 <0.11.0

### LINKED ISSUE #108: Compatibility issue with the latest video_player 0.10.0
Because chewie 0.9.3 depends on video_player >=0.7.0 <0.9.0 and no versions of chewie match >0.9.3 <0.10.0, chewie ^0.9.3 requires video_player >=0.7.0 <0.9.0.

So, because ddd depends on both chewie ^0.9.3 and video_player ^0.10.0, version solving failed.

Sub6Resources:
Fixes #108

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -11,7 +11,7 @@ environment:
 
 dependencies:
   open_iconic_flutter: ">=0.3.0 <0.4.0"
-  video_player: ">=0.7.0 <0.9.0"
+  video_player: ">=0.10.0 <0.11.0"
   screen: ">=0.0.4 <0.1.0"
   flutter:
     sdk: flutter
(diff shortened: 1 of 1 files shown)


hamidabdulmalik:
Isn't this taking too long?

cbenhagen:
I am preparing a release with this change. If you need this faster it's very easy to just checkout the code, change the dependencies and use this local version in your `pubspec.yaml`.

Closing this PR as I prefer to keep the dependencies wider than that. `">=0.7.0 <0.10.0"` will not force users to update to the new video_player and we do not depend on anything that changed in 0.10.0.

brianegan:
@hamidabdulmalik With all respect, please think a bit more before leaving similar comments in the future. We are giving away this software for free and we do it for the love of the community during our free time.

Comments like yours really hurt that sense of community.

PR #112, opened 2019-02-10
ChewieController.customControls with lazy function support

theafien:
this pull request add support to this scene: 

customControls is a class widget with access to controller

```
        aspectRatio: 16 / 9,
        allowFullScreen: true,
        autoPlay: true,
        allowMuting: false,
        customControls: _buildChewieCustomControls,
        videoPlayerController: _videoPlayerController);

  }

  Widget _buildChewieCustomControls(ChewieController controller) {
    return ChewieCustomControls(controller: controller); // this widget have access to controller
  }
```

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -193,7 +193,7 @@ class ChewieController extends ChangeNotifier {
 
   /// Defines customised controls. Check [MaterialControls] or
   /// [CupertinoControls] for reference.
-  final Widget customControls;
+  final dynamic customControls;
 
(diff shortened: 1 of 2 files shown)


brianegan:
Hey there -- I don't think this should be quite necessary! Your Custom controls can access the `ChewieController` via `ChewieController.of(context)` inside the build method! This is how the `Cupertino` and `Material` controls work already :) 

https://github.com/brianegan/chewie/blob/master/lib/src/cupertino_controls.dart#L45

Issue #113, opened 2019-02-11
Progress bar goes out of the interface when playing live resources


feilongfl:
![photo_2019-02-11_12-26-09](https://user-images.githubusercontent.com/10269596/52546223-7041a380-2df8-11e9-8173-64c493451550.jpg)

Progress bar goes out of the interface when playing live resources.

feilongfl:
I fix this issue and make a pr for you at https://github.com/brianegan/chewie/pull/115

brianegan:
Thanks much!

PR #115, opened 2019-02-13
fix progress bar goes out while play live stream.

feilongfl:
fix this issue: 
https://github.com/brianegan/chewie/issues/113


brianegan:
Excellent, thanks :)

PR #149, opened 2019-05-03
Fix rounded corners of iOS controls (remove transparent background - using ClipRRect)

Eimji:
Hello

Thanks for the package. There is a rendering bug with the rounded corners of iOS controls (the corners are rounded but there are transparent background below). So my PR is to fix this issue, using the ClipRRect class.


localpcguy:
@brianegan or @cbenhagen - any chance we can get this reviewed & merged if approved? It's a pretty simple fix, and I've been using it locally and it seems to work well.

brianegan:
Thanks!

Issue #125, opened 2019-02-28
Cannot exit fullscreen programmatically


MichaelFerrier:
I create a chewie player that defaults to fullscreen. I register a listener with the VideoPlayerController, and use it to exit fullscreen automatically once the video has nearly finished. However, exiting fullscreen doesn't work. The controls and the status bar briefly appear, then disappear again, and the video remains fullscreen. It does this repeatedly because chewieController.isFullScreen remains true, so it keeps calling exitFullScreen() repeatedly. 

Here's how I set up the ChewieController:
    
```
chewieController = ChewieController(
            videoPlayerController: videoPlayerController,
            aspectRatio: 16 / 9,
            autoPlay: true,
            fullScreenByDefault: true,
            looping: false,
          );
```

Here's the VideoPlayerController's listener:

```
static final VoidCallback videoPlayerListener = ()
  {
    if ((chewieController != null) && (videoPlayerController != null) && (chewieController.isFullScreen) && (videoPlayerController.value.position >= Duration(seconds: 234))) {
        chewieController.exitFullScreen();
      }
    }
  };
```

deche:
I'm also having the same issue. Can't exit fullscreen if fullScreenByDefault is set to true.

MichaelFerrier:
Thanks deche for pointing out that fullscreen can be exited if fullScreenByDefault is false. That lets me get the result I want, by setting fullScreenByDefault to false and then having the listener call enterFullScreen() at the start of the video. (Still would be nice if it worked correctly though :-) )

deche:
Glad that helped. I too finally used a listener, call enterFullScreen() the first time video plays.

lodemo:
Im also having issues with exiting fullscreen.

I have a listener which should exit fullscreen when the video is finished. The exitFullScreen call is executed, but does not change the state of the fullscreen.

``` 
_videoController.addListener(() {
  if ((_chewieController.isFullScreen) && (_videoController.value.position >= _videoController.value.duration)) {
        _chewieController.exitFullScreen();
      }
});
```

The fullscreen is not exited, and even by manually taping on the fullscreen control nothing changes. Back button press exits the fullscreen normally.

After looking in the Chewie code, exitFullScreen is setting the _isFullScreen false and then notifies the listeners.

```  
void exitFullScreen() {
    _isFullScreen = false;
    notifyListeners();
}
```

[...]

PR #152, opened 2019-05-13
fixes #132: only check fullScreenByDefault once

### LINKED ISSUE #132: FullScreenByDefault forces fullscreen after exiting fullscreen
FullScreenByDefault forces fullscreen after exiting fullscreen.
The api states that it should only define how it starts:
Defines if the player will start in fullscreen when play is pressed

So basicly when i press the button to exit fullscreen mode, it goes immediately back to fullscreen mode.

```    if (fullScreenByDefault) {
      videoPlayerController.addListener(() async {
        if (videoPlayerController.value.isPlaying && !_isFullScreen) {
          enterFullScreen();
        }
      });
    }
```
I think this code here is the culprit, when i comment it out everything works fine.

My settings are:

```
    chewieController = ChewieController(
      videoPlayerController: widget.controller,
      aspectRatio: 16 / 9,
      autoInitialize: true,
      fullScreenByDefault: true,
      looping: true,
      autoPlay: true
    );
```

Maybe the fault is on my side?

localpcguy:
updated the fullScreenByDefault logic to check on initialization and enter fullScreen and then remove the Listener after executing once. This prevents the listener from firing again on user action.

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -302,11 +302,14 @@ class ChewieController extends ChangeNotifier {
     }
 
     if (fullScreenByDefault) {
-      videoPlayerController.addListener(() async {
-        if (videoPlayerController.value.isPlaying && !_isFullScreen) {
-          enterFullScreen();
-        }
-      });
+      videoPlayerController.addListener(_fullScreenListener);
+    }
+  }
+
+  void _fullScreenListener() async {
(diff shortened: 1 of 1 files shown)


localpcguy:
This should also fix #125

brianegan:
Excellent, thanks so much :)

PR #172, opened 2019-06-27
Update README

MillerAdulu:
Adds the package video_player to the dependencies file. From a superficial point of view, it seems as if the chewie package comes with video_player hence it can be confusing when trying to instantiate `VideoPlayerController` and there's an error.


brianegan:
Thanks!

PR #189, opened 2019-08-27
fixes #88 - Replay on play when video finished

### LINKED ISSUE #88: Replay button on video finish
Hello,
i think it would make sense to display either a replay button when the video finished (and looping is false), or reset the video to the start and display the play button again. 
At the moment, at least for me, the video just ends and we have to manually slide the progress indicator back to the start if you want to watch it again.

VictorUvarov:
Fixes the issue with pressing the play button when the video is finished. Before the video player would keep playing even if the video has finished. 

This pr relates to what [#88](https://github.com/brianegan/chewie/issues/88) suggested but does not add a replay button icon.


brianegan:
Thanks, looks good!

PR #210, opened 2019-10-04
Switch wakelock plugin

### LINKED ISSUE #141: Play video cannot save the screen from always waking up
Play a long video, the screen won't stay awake, if you don’t touch the screen, it will lock the screen with the sleep time set by the system. If I want to keep the screen awake all the time, what should I do.

creativecreatorormaybenot:
The current wakelock plugin (`flutter_screen`) lacks maintenance and has had AndroidX compatibility issues for over half a year.

To fix this, I changed the plugin to `wakelock`, which is also more lightweight because it has just the wakelock and no brightness adjustments.

The formatting `cupertino_controls.dart` is caused by `flutter format` and Pub complained about it before.

The version constraint is correct like this and I will ensure that there are no compatibility changes until `0.2.0` as I am the author of [`wakelock`](https://pub.dev/packages/wakelock).

 * Closes #141 (that one is already solved, but including it in this comment will automatically close it upon merging).

## Related

https://github.com/clovisnicolas/flutter_screen/pull/15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.10.0
+
+* Changed wakelock plugin from `flutter_screen` to `wakelock` due to lack of maintenance of `flutter_screen`. 
+
 ## 0.9.8+1
   * Require latest flutter stable version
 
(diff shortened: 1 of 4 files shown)


brianegan:
Awesome, thanks for the update!

PR #214, opened 2019-10-16
Fixed setState() called after dispose() on ProgressBarState and Contr…

ilyasKerbal:
I am working on an app that has the ability to play videos from both online/offline. 

1. Online: VideoPlayerController.network
2. Offline: VideoPlayerController.file 

When switching from offline to online I've got this weird error:
`setState() called after dispose(): _VideoProgressBarState#51ebd(lifecycle state: defunct, not mounted)`

**Solution:**  Override setState method and check if mounted before updating sate

### DIFF
--- a/lib/src/cupertino_controls.dart
+++ b/lib/src/cupertino_controls.dart
@@ -39,6 +39,13 @@ class _CupertinoControlsState extends State<CupertinoControls> {
   ChewieController chewieController;
 
   @override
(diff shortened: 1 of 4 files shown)


Ahmadre:
@ilyasKerbal Thank you for this solution! I also conditioned the _updateState(), but this solution is much better :)

brianegan:
Thanks!

PR #222, opened 2019-10-25
增加alignment字段控制视频排列

zhangweiwei0326:
由于placeholder 能够居左，视频默认是center 此处改成Container 便于自定义

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -169,6 +169,7 @@ class ChewieController extends ChangeNotifier {
   ChewieController({
     this.videoPlayerController,
     this.aspectRatio,
+    this.alignment,
     this.autoInitialize = false,
     this.autoPlay = false,
     this.startAt,
@@ -233,6 +234,10 @@ class ChewieController extends ChangeNotifier {
   ///
   /// Will fallback to fitting within the space allowed.
   final double aspectRatio;
+  
+  /// video alignment
+  /// default Alignment.center
+  final Alignment alignment;
 
   /// The colors to use for controls on iOS. By default, the iOS player uses
   /// colors sampled from the original iOS 11 designs.
--- a/lib/src/player_with_controls.dart
+++ b/lib/src/player_with_controls.dart
@@ -32,7 +32,8 @@ class PlayerWithControls extends StatelessWidget {
       child: Stack(
         children: <Widget>[
           chewieController.placeholder ?? Container(),
-          Center(
+          Container(
+            alignment: chewieController.alignment ?? Alignment.center,
             child: AspectRatio(
               aspectRatio: chewieController.aspectRatio ??
                   _calculateAspectRatio(context),


Ahmadre:
> 由于placeholder 能够居左，视频默认是center 此处改成Container 便于自定义

Hello @zhangweiwei0326 ,

We would appreciate it, if you can talk with us in English :) otherwise we'll try to translate, but it could be misunderstood ^^. Anyways: For what use cases do you need alignment? Could you demonstrate it via a GIF, Code or something?

Best regards 
Rebar

syssam:
@Ahmadre 

The user can set placeholder as alignment left, but the video container must be center.
That's why he wants to add alignment.
However, I don't understand why placeholder use alignment left.
normally, the placeholder should be a video cover.

Ahmadre:
> @Ahmadre 
> 
> 
> 
> The user can set placeholder as alignment left, but the video container must be center.
> 
> That's why he wants to add alignment.
> 
> However, I don't understand why placeholder use alignment left.
> 
> normally, the placeholder should be a video cover.
> 
> 

Actually this hasn't to do with chewie but with `video_player`. But that issue what you're describing doesn't exist on video player. I always see the cover. And if you don't try to await initialize(); before you show any view.

Ortes:
Ported this onto current `master` (chewie 1.13.0 / Flutter 3.47) and tested it in Chrome.

**What it enables.** `player_with_controls.dart` hardcodes `Center` around the video, so when `aspectRatio` doesn't fill the space Chewie was given, the video is stuck in the middle. Below: `topLeft` at `aspectRatio: 1.0`, left = `master`, right = with this PR. Green = the space given to Chewie, yellow = the video's box.

before / after

**Why a caller can't do this from outside.** Wrapping `Chewie` in an `Align` moves the whole player — video, controls, scrim together — and the video stays centered *within* it. This PR moves the video alone: the controls remain a sibling `Stack` child spanning the full player box, as does the scrim, placeholder and overlay. Same `topLeft`, with `showControls: true` (magenta = the video's box, the frame didn't decode in this headless capture):

controls stay full-size

That decoupling is the point of the feature, but it's worth saying so in the doc comment — with `alignment` set, the controls no longer sit under the video.

Measured video box in a 1210x659 space, unpatched → patched: `topLeft`/ar1.0 `x=427→151`, `bottomRight`/ar3.0 `y=280→408`, `centerRight`/ar0.5 `→x=1031`, `center`/ar1.0 `x=427→427` (default unchanged). `flutter analyze lib` clean, `flutter test` 3/3.

**Two nits before merge:** `copyWith` (added since 2019) doesn't forward `alignment`, so it silently resets to center. And `Container(alignment:)` is `Align` with extra indirection — `Align(alignment: chewieController.alignment ?? Alignment.center)` is the direct swap for the `Center` it replaces, ideally typed `AlignmentGeometry` so RTL callers can pass `AlignmentDirectional`.

PR #225, opened 2019-10-31
Adds the possibility to set on full screen preferred orientations

miguelpruivo:
Currently developers are restricted to use landscape right or left when in full screen while using an Android device, whereas in iOS it will adapt automatically. This will result in an odd behavior when playing portrait (aspect ratio 9:16 for example) videos. 

This will add the `deviceOrientationsOnFullScreen` property that accepts a `List<Orientation>` for full screen playback that will be used for both platforms. If not provided, it will use de default landscape right/left for Android, as it is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.9.9
+  * Adds the possibility to set on full screen preferred screen orientations.
+
 ## 0.9.8+1
   * Require latest flutter stable version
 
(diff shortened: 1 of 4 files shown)


brianegan:
Thanks @miguelpruivo! Overall, I'm happy to make this change. One question, however: Why did you roll back the `Hero` changes? We removed those to fix some other issues, so it'd be great to hear if adding it back in fixed something for ya.

miguelpruivo:
Sorry @brianegan I forgot to cherry pick the commit to not use the `Hero` rollback on this PR. I’ve added it on my personal fork because I actually enjoyed the effect when entering/exiting from full screen, special when you play it along with scroll down gesture to dismiss, for example, like Facebook has on its pictures.

miguelpruivo:
Opened #226 from another source branch. I'm closing this now @brianegan.

PR #226, opened 2019-10-31
Adds the possibility to set on full screen preferred orientations

miguelpruivo:
Currently developers are restricted to use landscape right or left when in full screen while using an Android device, whereas in iOS it will adapt automatically. This will result in an odd behavior when playing portrait (aspect ratio 9:16 for example) videos.

This will add the `deviceOrientationsOnFullScreen` property that accepts a `List<Orientation>` for full screen playback that will be used for both platforms. If not provided, it will use de default landscape right/left for Android, as it is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.9.9
+  * Adds the possibility to set on full screen preferred screen orientations.
+
 ## 0.9.8+1
   * Require latest flutter stable version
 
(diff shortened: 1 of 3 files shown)


miguelpruivo:
@brianegan regarding the comment in #225 about the `Hero` widget, as a suggestion, it would be great if the `Hero` could be optionally enabled instead of removing at all.

Ahmadre:
@miguelpruivo applied your wishes in this PR: https://github.com/brianegan/chewie/pull/289 like @brianegan wanted to be in that PR. Please follow that PR for onEnterFullScreen :)

miguelpruivo:
That's great news @Ahmadre, thank you!

alexxlagutin:
Is there any information when this PR will be merged?

Ahmadre:
@miguelpruivo @alexxlagutin Since version 0.10.x it's possible :) a big pr and many other changes came in 🥳

PR #235, opened 2019-11-15
User can now set autoPlay to false

ouardia-y:
fixed issue with setting autoPlay to false. The user can now set autoPlay to false and the video will still be loaded. The issue was that durant was null and couldn't render the video. Another future fix would be to display the play button instead of the loading icon when autoPlay is set to false.


chenzhengxu:
it works

p30arena:
the code seems to be ok

brianegan:
Hey, thanks so much!

Issue #239, opened 2019-11-23
FullScreen video rebuilding MaterialApp and cause crashs


jonataslaw:
When some users start full screen video they are crashing due to unexpected MaterialApp rebuild because the plugin uses opaque routes.

djsjr:
Why was the PR never merged or fixed? Any update on this huge issue? A workaround at least?

PR #240, opened 2019-11-23
Prevent rebuild of MaterialApp

jonataslaw:
PageRouteBuilder need "opaque: false," to prevent unexpected MaterialApp rebuild. 
This commit fix https://github.com/brianegan/chewie/issues/239


searchy2:
This improves performance a lot especially if the Chewie player is backed in a ListView.

PR #243, opened 2019-12-10
add an "isPlaying" property and a "togglePause" method

CoderMonkie:
add an "isPlaying" property and "togglePause" method


brianegan:
Thanks!

PR #247, opened 2019-12-16
Add rtl

mreslami:
Automatic rtl detection and field rtl in chewieController for the display buildBottomBar

### DIFF
--- a/lib/src/material_controls.dart
+++ b/lib/src/material_controls.dart
@@ -6,7 +6,6 @@ import 'package:chewie/src/material_progress_bar.dart';
 import 'package:chewie/src/utils.dart';
 import 'package:flutter/material.dart';
 import 'package:video_player/video_player.dart';
-
 class MaterialControls extends StatefulWidget {
   const MaterialControls({Key key}) : super(key: key);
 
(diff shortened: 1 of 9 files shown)


lib/src/material_controls.dart
```diff
@@ -108,32 +107,39 @@ class _MaterialControlsState extends State<MaterialControls> {
...
-          ],
+          children: createWidget(iconColor)
         ),
       ),
     );
   }
 
+ List<Widget> createWidget(Color iconColor){
+   Locale myLocale = Localizations.localeOf(context);
+   isRtlLanguage( myLocale.languageCode);
```
nstrelow, line 125: Unneeded line

nstrelow:
Could you explain why this is needed?
Does the video player bottom bar need to be inverted for rtl countries?

Your solution is well thought out but misses a few things.

In my opinion we should not include localization libraries just to find out if rlt and then reverse the controls.

Is there any benefit or real need to reverse the controls?

@Ahmadre What's your opinion?

Ahmadre:
> Could you explain why this is needed?
> 
> Does the video player bottom bar need to be inverted for rtl countries?
> 
> 
> 
> Your solution is well thought out but misses a few things.
> 
> 
> 
> In my opinion we should not include localization libraries just to find out if rlt and then reverse the controls.
> 
> 
> 
> Is there any benefit or real need to reverse the controls?
> 
> 
> 
> @Ahmadre What's your opinion?

Rtl is not only related to localizations, but also to accessibility. It would be nice if left-handed people enjoy apps made with chewie.

So, thumbs up from me for this 👍🏼

diegotori:
@mreslami please re-sync your changes with the latest master changes, then I'll be able to finish reviewing. Thanks in advance.

PR #252, opened 2019-12-30
[Fixed] Seeking after video play finished causes inconsistent player state.

rashedoz:
Sorry if there's any mistake in the pull request as this is my first time contributing. For this fix, only this files changes are needed https://github.com/rashedoz/chewie/blob/master/lib/src/material_progress_bar.dart . 

-----------------------
The other modifications are related to HLS bitrate change support for flutter video player. I have also modified flutter core video player to support HLS  bitrate changes. I have also created a custom chewie control to support bitrate(360p,380p,720p,1080p) change in HLS .
Screenshot_1577679911
Screenshot_1577679916


nstrelow:
Is this still an issue in the current version?

And could you please update your commits to only include relevant files?

I do not think we should change the video_player dep away from the officially supported one.

Thank you!

PR #260, opened 2020-01-14
[WIP] Crop to aspect ratio

cbenhagen:
As is, this would be a breaking change that automatically gets the aspect ratio when ready and use the current `AspectRatio` set in the `ChewieController` to crop the video. This is useful to crop letterboxes for example.

### DIFF
--- a/lib/src/player_with_controls.dart
+++ b/lib/src/player_with_controls.dart
@@ -32,12 +32,9 @@ class PlayerWithControls extends StatelessWidget {
       child: Stack(
         children: <Widget>[
           chewieController.placeholder ?? Container(),
-          Center(
-            child: AspectRatio(
-              aspectRatio: chewieController.aspectRatio ??
-                  _calculateAspectRatio(context),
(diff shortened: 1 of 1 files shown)


Ahmadre:
@cbenhagen when is this going to be merged?

AndroidNils:
That pr would help's me a lot!

Ahmadre:
@AndroidNils fixed AspectRatio, take a look here: https://github.com/brianegan/chewie/pull/289

brianegan:
Thanks so much :)

This is a tricky one, and had discussion with @Ahmadre and others on a different channel.  #289 seems to fix the issue for folks! Closing in favor of that one.

cbenhagen:
@brianegan https://github.com/brianegan/chewie/pull/289 does not seem to allow cropping a video. For example if you want to remove the letterboxes from a video.

Ahmadre:
@cbenhagen Sure! We definetly should also have a cropping functionality in chewie. Maybe as an option?

PR #274, opened 2020-02-19
Add functions before and after fullscreen

Ahmadre:


Ahmadre:
would fix this: https://github.com/brianegan/chewie/issues/280#issue-573546843

Ahmadre:
I encountered to check for null Functions, because this PR runs for now in errors and won't enter FullScteen. I will close this for now and add a new PR, because repos changed.

PR #289, opened 2020-03-24
Correct AspectRatio Issue | Allow Portrait

Ahmadre:
Thank you @brianegan and all other contributors for this great 📦

I saw some bugs in the player_with_controls and corrected the AspectRatio issue with Portrait Mode and enabled Portrait as of correcting it.

I tested it several times and now chewie is calculating the video aspect ratio correctly. Furthermore I am also able to set the custom aspectRatio in the parameter if needed. But due to this fix maybe it's useless for the future.

Result (inScreen):
Bildschirmfoto 2020-03-24 um 01 47 54

Result (FullScreen):
Bildschirmfoto 2020-03-24 um 01 58 57

The problem in player_with_controls were:
  1. the `_calculateAspectRatio` only calculated AspectRatio related to the device and is setting it for the whole view: Video + Controls. But the AspectRatio from the video was never used.
  2. The Height was ignored in the `Container` so it was necessary to set it also to enable Portrait correctly.


xsahil03x:
This is exactly what I needed. :+1:

brianegan:
Excellent, thanks so much @Ahmadre !

lib/src/chewie_player.dart
```diff
@@ -129,6 +129,7 @@ class ChewieState extends State<Chewie> {
     SystemChrome.setEnabledSystemUIOverlays([]);
     if (isAndroid) {
       SystemChrome.setPreferredOrientations([
+        DeviceOrientation.portraitUp,
```
brianegan: Is this required for the fix as well? If so, I think it might be good to completely remove these preferredOrientations as the only one missing would be portraitDown.

Ahmadre: Good point! i'll remove it and test it

Ahmadre: Do we need this:
https://github.com/brianegan/chewie/blob/02b3bf2f9d3bdb669c1bec92dcd5bac0ea508e03/lib/src/chewie_player.dart#L150
and this:
https://github.com/brianegan/chewie/blob/02b3bf2f9d3bdb669c1bec92dcd5bac0ea508e03/lib/src/chewie_player.dart#L152
anymore, if we remove PreferredOrientations and EnabledSystemUIOverlays?

brianegan: I think we'll definitely want to keep the `setEnabledSystemUIOverlays `, but the preferred orientations is trickier.

I could see taking a few approaches, would love input from you and others on what would work best for everyone:

  1. Keep the orientation lock, but make it smarter:  set the preferred orientations based on the aspect ratio of the video, e.g. PortraitUp/Down for h > w videos, and landscapeLeft/Right for w > height videos. This would still allow the "auto switch to landscape in fullscreen" which is common for Android devices, but would still work for portrait vids. This means we would keep the `setPreferredOrientations` line from above when exiting fullscreen.
  2. Allow folks to customize those orientations, as is done in this PR:. Also keeps the `setPreferredOrientations` line from above when exiting fullscreen.
  3. A combination of those two: Use the orientations defined by the user if provided, fallback to detecting the appropriate orientation lock if not. Also keeps the `setPreferredOrientations` line from above when exiting fullscreen.
  4. Remove those lines! The only downside: That would disable the "auto rotate to Landscape upon entering fullscreen" feature, which feels like the default for most android video players. In this case, we could remove the `setPreferredOrientations` line when exiting fullscreen.

Do you have an opinion on which of these three would be the best option for your use case? I only use Chewie for a few small things, so I'd love your feedback!

Ahmadre: Thank you again @brianegan for your time to discuss about the solution for this. 

I will summarize it here in short:

1. We'll have two new parameters, that users can set and have default values:
```dart
    this.systemOverlaysOnEnterFullScreen = SystemUiOverlay.values,
    this.deviceOrientationsOnEnterFullScreen = DeviceOrientation.values,
``` 
2. Chewie will set the `setPreferredOrientations` based on the videos aspectRatio (h>w = Portrait and w>h = Landscape) for both use cases: OnEnterFullScreen and AfterFullScreen. Is that correct? Or just for OnEnterFullScreen?

Thanks for your feedback :)

Ahmadre:
@brianegan I've done all changes and tested it with emulator and my physical devices (Pixel 2XL, Ipad 6G).

Following changes in short: 
  * Fix portrait mode
  * Add auto-detect orientation based on video aspect-ratio
  * Add optional parameters for `onEnterFullScreen`

I also replaced the GIF with a new one like the old one, but now with new videos and representation of landscape and portrait videos. 

I leaved out the optional new parameters (`systemOverlaysOnEnterFullScreen` and `deviceOrientationsOnEnterFullScreen`) and the default behavior was set and the expected behavior depending on videos aspect-ratio was set. So the combination of 1 + 2, so 3 is now implemented 👍. 

Therefore the PR: https://github.com/brianegan/chewie/pull/226 is hearby also solved.

Furthermore I removed the aspectRatio hint in the readme examples and added `initialize()` method, so users won't get in aspectRatio issues anymore and get on default always the right aspectRatio and expected orientation 👍  

I already added a new note in changelog and raised the version in pubspec.yaml, if it's ok :).

Wish you all a nice weekend ✌️

Ahmadre:
@brianegan did you had some time to look into this? Is it ok?

Dreamersoul:
Yeah I made it work! Thank you so much for the help!

Ahmadre:
For all who don't want to wait for the PR to land, add these lines to your pubspec.yaml:

```yaml
chewie:
  git:
    url: git://github.com/Ahmadre/chewie.git
    ref: Correct-Portrait
```

daohoangson:
Hey @Ahmadre, I think your PR is making it impossible to use Chewie with other content in the same screen. Is that correct? Here is the same code before and after this PR being merged:

| Before | After |
| --- | --- |
| Screenshot_20201114-151237 | Screenshot_20201114-150944 |

I think this change broke it:

```diff
-          aspectRatio: chewieController.aspectRatio ?? _calculateAspectRatio(context),
+          aspectRatio: _calculateAspectRatio(context),
```

Is it possible to revert this? And maybe the fixed height/width?

Ahmadre:
> Hey @Ahmadre, I think your PR is making it impossible to use Chewie with other content in the same screen. Is that correct? Here is the same code before and after this PR being merged:
> 
> 
> 
> | Before | After |
> 
> | --- | --- |
> 
> | Screenshot_20201114-151237 | Screenshot_20201114-150944 |
> 
> 
> 
> I think this change broke it:
> 
> 
> 
> ```diff
> 
> -          aspectRatio: chewieController.aspectRatio ?? _calculateAspectRatio(context),
> 
> +          aspectRatio: _calculateAspectRatio(context),
> 
> ```
> 
> 
> 
> Is it possible to revert this? And maybe the fixed height/width?

You're doing something wrong here. This PR brought several changes which were also discussed in a live meeting with Mr. Egan. So no on my side your issue is not appearing.

Please share a reproducable code that I can investigate on this, then I'll double check everything.

Thanks 🙏🏼

daohoangson:
Please find the reproduce code below.

<details>

<summary>video_screen.dart</summary>

```dart
import 'package:flutter/material.dart';
import 'package:chewie/chewie.dart';
import 'package:video_player/video_player.dart';

class VideoScreen extends StatefulWidget {
  @override
  State<StatefulWidget> createState() => _State();
}

class _State extends State<VideoScreen> {
  final videoPlayerController = VideoPlayerController.network(
      'https://flutter.github.io/assets-for-api-docs/assets/videos/butterfly.mp4');
  ChewieController chewieController;

  @override
  void initState() {
    super.initState();
    chewieController = ChewieController(
      aspectRatio: 16 / 9,
      autoInitialize: true,
      videoPlayerController: videoPlayerController,
    );
  }

  @override
  void dispose() {
    chewieController.dispose();
    videoPlayerController.dispose();
    super.dispose();
  }

  @override
  Widget build(BuildContext context) => Scaffold(
        appBar: AppBar(
          title: Text('VideoScreen'),
        ),
        body: ListView(children: <Widget>[
          Text('Above'),
          Chewie(controller: chewieController),
          Text('Below'),
        ]),
      );
}
```

</details>

The before screenshot was with chewie v0.9, the after was with v0.10. I have also tried v0.12, it's similar to v0.10.

Ahmadre:
> Please find the reproduce code below.
> 
> 
> 
> <details>
> 
> 
> 
> <summary>video_screen.dart</summary>
> 
> 
> 
> ```dart
> 
> import 'package:flutter/material.dart';
> 
> import 'package:chewie/chewie.dart';
> 
> import 'package:video_player/video_player.dart';
> 
> 
> 
> class VideoScreen extends StatefulWidget {
> 
>   @override
> 
>   State<StatefulWidget> createState() => _State();
> 
> }
> 
> 
> 
> class _State extends State<VideoScreen> {
> 
>   final videoPlayerController = VideoPlayerController.network(
> 
>       'https://flutter.github.io/assets-for-api-docs/assets/videos/butterfly.mp4');
> 
>   ChewieController chewieController;
> 
> 
> 
>   @override
> 
>   void initState() {
> 
>     super.initState();
> 
>     chewieController = ChewieController(
> 
>       aspectRatio: 16 / 9,
> 
>       autoInitialize: true,
> 
>       videoPlayerController: videoPlayerController,
> 
>     );
> 
>   }
> 
> 
> 
>   @override
> 
>   void dispose() {
> 
>     chewieController.dispose();
> 
>     videoPlayerController.dispose();
> 
>     super.dispose();
> 
>   }
> 
> 
> 
>   @override
> 
>   Widget build(BuildContext context) => Scaffold(
> 
>         appBar: AppBar(
> 
>           title: Text('VideoScreen'),
> 
>         ),
> 
>         body: ListView(children: <Widget>[
> 
>           Text('Above'),
> 
>           Chewie(controller: chewieController),
> 
>           Text('Below'),
> 
>         ]),
> 
>       );
> 
> }
> 
> ```
> 
> 
> 
> </details>
> 
> 
> 
> The before screenshot was with chewie v0.9, the after was with v0.10. I have also tried v0.12, it's similar to v0.10.
> 
> 

Jup, i see what is missing here: Your video aspect-ratio was never calculated, because you never do: `initialize()` on your videoPlayerController. And I think this lead to expanding in the height. Did you use autoInitialize parameter of chewie?

daohoangson:
The reproduce code has `autoInitialize: true`, doesn't it initialize the underlying player by itself?

daohoangson:
So I modified the code to wait for `VideoPlayerController.initialize` before creating the `ChewieController` but it's showing up in wrong aspect ratio... 

<details>

<summary>video_screen.dart</summary>

```dart
import 'dart:async';

import 'package:flutter/material.dart';
import 'package:chewie/chewie.dart';
import 'package:video_player/video_player.dart';

class VideoScreen extends StatefulWidget {
  @override
  State<StatefulWidget> createState() => _State();
}

class _State extends State<VideoScreen> {
  final videoPlayerController = VideoPlayerController.network(
      'https://flutter.github.io/assets-for-api-docs/assets/videos/butterfly.mp4');
  final chewieController = Completer<ChewieController>();

  @override
  void initState() {
    super.initState();

    videoPlayerController
        .initialize()
        .then((_) => chewieController.complete(ChewieController(
              aspectRatio: 16 / 9,
              autoInitialize: true,
              videoPlayerController: videoPlayerController,
            )));
  }

  @override
  void dispose() {
    videoPlayerController.dispose();
    super.dispose();
  }

  @override
  Widget build(BuildContext context) => Scaffold(
        appBar: AppBar(
          title: Text('VideoScreen'),
        ),
        body: ListView(children: <Widget>[
          Text('Above'),
          FutureBuilder(
            builder: (_, snapshot) => snapshot.hasData
                ? Chewie(controller: snapshot.data)
                : Center(child: CircularProgressIndicator()),
            future: chewieController.future,
          ),
          Text('Below'),
        ]),
      );
}
```

</details>

Check this out:

daohoangson:
What do you think @Ahmadre?

cdinarte:
I'm experiencing the same problem reported by @daohoangson. Downgrading chewie version actually fixes the problem.

```
_videoPlayerController = VideoPlayerController.network(value.videoUri.toString());
await _videoPlayerController.initialize();
  aspectRatio: _videoPlayerController.value.aspectRatio
);
```

fbatschi:
Same problem as @daohoangson and @cdinarte , fixed it by wrapping with an AspectRatio Widget set to the videos aspect ratio.

andrezanna:
Same issue, resolved as pointed by @fbatschi. Many thanks

PR #290, opened 2020-03-28
problem in return button

MpTanin:
When the orientation of the page changes , the return button does not work .


brianegan:
Hey there -- was this PR made against the correct github repository? This code seems to be based "Custom Chewie" -- a fork of this library -- instead of Chewie (this library).

PR #300, opened 2020-04-21
Support for iOS Fullscreen

dhalloop7:
Removed condition for android only.
Working for iOS devices as well.

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -127,12 +127,12 @@ class ChewieState extends State<Chewie> {
     );
 
     SystemChrome.setEnabledSystemUIOverlays([]);
-    if (isAndroid) {
+//    if (isAndroid) {
       SystemChrome.setPreferredOrientations([
         DeviceOrientation.landscapeLeft,
         DeviceOrientation.landscapeRight,
       ]);
-    }
+//    }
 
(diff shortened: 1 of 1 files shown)


lib/src/chewie_player.dart
```diff
@@ -127,12 +127,12 @@ class ChewieState extends State<Chewie> {
     );
 
     SystemChrome.setEnabledSystemUIOverlays([]);
-    if (isAndroid) {
+//    if (isAndroid) {
```
AndroidNils, line 130: Why do you comment it out instead of removing this lines?

lib/src/chewie_player.dart
```diff
@@ -127,12 +127,12 @@ class ChewieState extends State<Chewie> {
     );
 
     SystemChrome.setEnabledSystemUIOverlays([]);
-    if (isAndroid) {
```
AndroidNils, line 130: What was the reason for only supporting this feature on Android?

w3ggy:
Could you fix this MR and merge it into the repo?

roskimlong:
Work well on the video link, but not work on live link :(

Ahmadre:
Hey guys 😊 This issue is already fixed with a big PR a day ago. Enjoy cross-platform 👌🏽

PR #309, opened 2020-05-13
Check widget is mounted

wolcy97:
When switching to another widget; the video continued to play on the back and wants to report transactions, but the widget failed because it was not mounted.
For this reason, it is provided to check whether the widget is mounted or not.


Ahmadre:
I encountered the same error. I'll check this with the latest version and let you know. Thanks @wolcy97 !

Ahmadre:
Erro still exists and the check in cupertino_controls and material_progress_bar was missing. Thanks for this PR!

PR #367, opened 2020-09-27
Add support for video play speed change control

plateaukao:


dgilperez:
+1, chewie dependency is preventing us to update `video_player` to 0.11. In our case, this issue comes from our dependency to https://github.com/daohoangson/flutter_widget_from_html/pull/330, that can't be upgraded either.

plateaukao:
Hi, @dgilperez 
if you just want to upgrade to `video_player` while using flutter_widget_from_html, you can just add following snippet into your pubspec.yml:

```yml
dependency_overrides:
  video_player: ^0.11.0
```

Ahmadre:
@plateaukao I've checked out your branch and tested it. But the speed is just working for MaterialControls and not for CupertinoControls also. And what I encountered is, that you tap on it and the speed changes. How about a DropDownButton with a list of available speed to select from :)? That would be great for this PR. Here my test:

example

Ahmadre:
Thanks @plateaukao for your work, there came a PR in, which got both controls. Please follow that PR: https://github.com/brianegan/chewie/pull/390

PR #378, opened 2020-10-15
Fix video url in the demo

KarimMohamed20:
The video url at the line 26 wasn't working


Ahmadre:
Already fixed. New Video (GIF actually) is uploaded and we're on road again :)

PR #379, opened 2020-10-15
Change the video url in demo

KarimMohamed20:
The video that chewie using was deleted by his owner
And I added a new one


Ahmadre:
Already fixed. New Video (GIF actually) is uploaded and we're on road again :)

PR #382, opened 2020-10-27
Remove iconic font for improved App Store compatibility

### LINKED ISSUE #381: iOS apps with iOS 14 widget and chewie will get rejected on ITMS-90853: Invalid font
If you add chewie to an iOS app that also has an iOS 14 widget (extension) and submit this for (TestFlight) review, it will get rejected:

> **ITMS-90853**: Invalid font - There was an error validating the font at 'Runner.app/Frameworks/App.framework/flutter_assets/packages/open_iconic_flutter/assets/open-iconic.woff'. Font not supported by platform. 

This is due to the dependency on `open_iconic_flutter` which embeds the open-iconic font in unsupported woff format. 

I am not sure why adding an iOS widget triggers this static analysis at Apple's side, but it is somewhat widely reported that it does, similar to how it did [in Flutter itself](https://github.com/flutter/flutter/issues/65991).

erickok:
Fixes #381 by removing the `open_iconic_flutter` dependency (which was bringing in the unsupported woff font file), replacing the icons with Cupertino (and Material, if appropriate) variants.

While not a pixel for pixel replacement, they are like for like as the iconography is identical.

Existing Iconic version:
iconic

Version of this PR, using Cupertino icons:
cupertino


diegotori:
@Ahmadre gentle reminder regarding this PR. I quickly reviewed it myself, and it looks good enough for approval IMHO.

Ahmadre:
@erickok thanks a lot! And @diegotori I'll check this ASAP today and merge everything 💪🏼

Ahmadre:
LGTM. @nstrelow what do you think?

Ahmadre:
@erickok is it possible to have the "15" inside the rewind and fast forward icons? Or: is there the exact same icon in cupertino 1.0.0 ?

Ahmadre (approved):

Ahmadre:
@erickok Could you also have a look in the merge conflicts in pubspec.yaml? I think just need to check the video player dependecy. It should be as same as the actual one in master, because an important fix came in 👍

nstrelow:
Just had a quick look, will have a deeper look later today.

- Awesome change, removing unneeded dependencies is always great, especially if it causes incompatibilities
- The code was reformatted in some places. That makes it harder to see the actual changes. I will setup a linting and formatting github action, that will check if formatting and linting is correct. Should make it easier to have consistent code.

Ahmadre:
> Just had a quick look, will have a deeper look later today.
> 
> * Awesome change, removing unneeded dependencies is always great, especially if it causes incompatibilities
> * The code was reformatted in some places. That makes it harder to see the actual changes. I will setup a linting and formatting github action, that will check if formatting and linting is correct. Should make it easier to have consistent code.

Love you man!!

lib/src/cupertino_progress_bar.dart
```diff
@@ -25,29 +25,10 @@ class CupertinoVideoProgressBar extends StatefulWidget {
...
-
-  VoidCallback listener;
   bool _controllerWasPlaying = false;
 
   VideoPlayerController get controller => widget.controller;
 
-  @override
-  void initState() {
-    super.initState();
-    controller.addListener(listener);
```
nstrelow, line 42: Why is the listener stuff not needed anymore?

nstrelow:
@Ahmadre Speaking of the forward/backward icons, why would you like a 15 in them? Is it like this on android?
There is a cuptertino icon called: goforward_15

Ahmadre:
> @Ahmadre Speaking of the forward/backward icons, why would you like a 15 in them? Is it like this on android?
> 
> There is a cuptertino icon called: goforward_15
> 
> 
> 
> See https://flutter.github.io/cupertino_icons/

Great :) then Cupertino is supporting it 👍🏼

I got a Pixel 2XL and an IPhone 11 Pro as test devices and on IPhone you got these controls in the native player:

image

It would fit perfect for iOS 14.

lib/src/cupertino_progress_bar.dart
erickok: This was never used (and incorrectly never unregistered).

nstrelow:
Oh yeah, I would love to use these. Especially **not** rotating the icons, if there are already icons that are rotated. Reduces complexity.

Do we acutally skip 15 back and forth? Then the 15 icons is perfect.
But we can do that in a different small PR and merge this one first.

lib/src/cupertino_progress_bar.dart
nstrelow: @Ahmadre or @erickok since I am not sure :)

erickok:
I guess an icon with '15' in it is possible, though in our player the icon is really quite small to cater for these numbers, so I doubt whether the legibility would be on par. Practically, though, I opted to reflect the existing chewie design to reduce the impact on existing chewie users.

Sorry for the formatting changes. dartfmt changes from time to time (or perhaps the code was never formatted with it).

Ahmadre:
> Oh yeah, I would love to use these. Especially **not** rotating the icons, if there are already icons that are rotated. Reduces complexity.
> 
> 
> 
> Do we acutally skip 15 back and forth? Then the 15 icons is perfect.
> 
> But we can do that in a different small PR and merge this one first.
> 
> 

Sure! This PR is very important that users can publish their iOS versions. Let'a merge this and do a feature PR on the "15". and no on iOS they're not rotating. no need for animation here :). By the way I am working on a feature branch for Animated Play/Pause Button and Replay Button. But that's for the future.

nstrelow:
@erickok Great, nothing holding it back from my side.

Created a new issue to check with different icons #387 

@Ahmadre Feel free to merge 🚀

Ahmadre:
@erickok I'll merge your urgent changes now and you can work on the new icons here: https://flutter.github.io/cupertino_icons/ like @nstrelow mentioned. New version is coming after I've tested chewie again (maybe we can have a CI here 👀)

PR #383, opened 2020-10-29
video_player version update

vkpdeveloper:
Hello Dear, I have added the latest video_platyer version in pubspec.yaml because it is not working with the latest video_player version because as your restriction for video_player version.


Ahmadre:
Already updated: https://github.com/brianegan/chewie/pull/386

Thanks anyway :)

PR #393, opened 2020-11-09
Add Linter

nstrelow:
Just FYI, added the lint package, this is how it would look like applying the changes.

I am only concerned, that if we merge this OR the CI changes with dartfmt line length 120, all current PRs need to be updated.

I would probably do it myself, so the kind contributors do not have to apply it themselves.

But therefore I would wait until we merge a few more, e.g. the video playback speed and then require the linting and formatting to pass.


analysis_options.yaml
```diff
@@ -1,18 +1,12 @@
...
-    - hash_and_equals
-    - iterable_contains_unrelated_type
-    - list_remove_unrelated_type
-    - sort_constructors_first
-    - test_types_in_equals
-    - unnecessary_new
-    - unrelated_type_equality_checks
-    - valid_regexps
+    # const is automatically added by the Dart VM
+    prefer_const_constructors: false
```
Ahmadre, line 10: Very good!

Ahmadre (approved):

Ahmadre:
Very good job @nstrelow 👌🏼

nstrelow:
Created an issue exploring the const removal:
https://github.com/passsy/dart-lint/issues/22

nstrelow:
@Ahmadre FYI This is merged now.

Maybe we should add a contributing.md or something with the style requirements.

Ahmadre:
> @Ahmadre FYI This is merged now.
> 
> Maybe we should add a contributing.md or something with the style requirements.

yes I saw that, good job 💪🏼 i agree

analysis_options.yaml
passsy: Very bad! https://github.com/passsy/dart-lint/issues/22#issuecomment-726751958

nstrelow: @passsy Thanks for having a look :P

Do you think const will provide a noticeable performance benefit for flutter apps?
If not I would rather not pollute the code with const

passsy: If you'd remove all `const` from the Flutter SDK you'd definitely notice it. A single package won't have a big impact on performance. But when we all push the usage of `const` it will make a difference.
Adding a few `const` here and there is a rather small price to pay.

nstrelow: Hmm, good point having it in packages will help not that being a bottleneck in an app implementing it.

PR #395, opened 2020-11-11
Fix dependencies

creativecreatorormaybenot:
@Ahmadre Just want to point out that it is not guaranteed that there will be no breaking changes until `1.0.0`. The Pub versioning would allow breaking changes when upgrading any **x** version in `0.x.y+z` or `x.y.z` when `x > 1`.

It is correct that there was no breaking change for `wakelock 0.2.0` ([changelog](https://pub.dev/packages/wakelock/changelog#020)) - only deprecations. However, I would probably want to remove the deprecations at some point.

---

About the `cupertino_icons` change: the caret syntax is identical in that case and the preferred syntax. See https://dart.dev/tools/pub/dependencies#caret-syntax.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,7 +8,7 @@ environment:
   flutter: ">=1.20.0 <2.0.0"
 
 dependencies:
-  video_player: ">=0.10.12+5 <2.0.0"
+  video_player: ^1.0.0
   cupertino_icons: ^1.0.0
   wakelock: ">=0.1.2 <0.3.0"
 
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -9,8 +9,8 @@ environment:
 
 dependencies:
   video_player: ">=0.10.12+5 <2.0.0"
-  cupertino_icons: ">=1.0.0 <2.0.0"
-  wakelock: ">=0.1.2 <1.0.0"
+  cupertino_icons: ^1.0.0
+  wakelock: ">=0.1.2 <0.3.0"
 
   flutter:
     sdk: flutter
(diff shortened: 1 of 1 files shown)


nstrelow:
Nice, what about also setting videoplayer to minimum of 0.11 ? 
to have the speed stuff?

https://pub.dev/packages/video_player/changelog#0110

creativecreatorormaybenot:
@nstrelow Good point 👍 We can actually use `^1.0.0` now for video player as well (given how the dependencies allowed that previously anyway).

Ahmadre:
@creativecreatorormaybenot sure let's do that but why did you downgrade wakelock? We need the latest wakelock for web support. After that this can be merged

nstrelow:
Would that break anybodies setup?
e.g. using a package, which restricts to before 1.0.0?

I would feel better if we generally specify the lowest supported number. But looking at all the fixes until 1.0.0, I am fine with making this a requirement

nstrelow:
@Ahmadre Newest wakelock is 0.2.1+1

According to semver.org everything **before** 1.0.0, so 0.x.y can include breaking changes.

Ahmadre:
> Would that break anybodies setup?
> 
> e.g. using a package, which restricts to before 1.0.0?
> 
> 
> 
> I would feel better if we generally specify the lowest supported number. But looking at all the fixes until 1.0.0, I am fine with making this a requirement

Yeah sure let's also do `^1.0.0` for wakelock :)

creativecreatorormaybenot:
The dependencies should be fine like this.

`wakelock` might go to `1.0.0` when we have support for all platforms and it looks like the API is in a good state.

nstrelow:
Great thanks!

Ahmadre:
> Great thanks!

Thank you guys for this merge 🙏

Ahmadre:
Alright guys, new version is online: [Version](https://pub.dev/packages/chewie)

Thanks all :)

PR #411, opened 2020-12-15
Add property to allow showing the controls in fullscreen

minhqdao:
Right now, 'showControls' only allows for enabling controls in both fullscreen and non-fullscreen mode. With the added 'showControlsInFullscreen' property, one can, e. g., show controls in fullscreen while hiding them in non-fullscreen mode.


syssam:
any news?

PR #437, opened 2021-02-15
Jeremiah-Parrack | [Chore] Fix lint

jeremiahlukus:
'FlatButton' is deprecated and shouldn't be used. Use TextButton instead. See the migration guide in flutter.dev/go/material-button-migration-guide). This feature was deprecated after v1.25.0-8.1.pre. •

### DIFF
--- a/example/ios/Flutter/Debug.xcconfig
+++ b/example/ios/Flutter/Debug.xcconfig
@@ -1,2 +1,3 @@
+#include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
 #include "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
 #include "Generated.xcconfig"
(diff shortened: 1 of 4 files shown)


example/ios/Flutter/Debug.xcconfig
```diff
@@ -1,2 +1,3 @@
+#include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
```
nstrelow, line 1: Are these new includes in example needed?

I don't own a Mac, so not sure if they are?

Also it's almost the same line as one beneath it. #include vs #include?

nstrelow:
Thanks so much for providing this change 👍 

If you could remove the change in ios files if not needed that would be great (or add the questionmark behind #include if that is new now)

example/ios/Flutter/Debug.xcconfig
jeremiahlukus: I have no idea it was auto generated so i figured it was needed, i can remove it though if you want 👍

nstrelow: Please remove it. Makes it a tiny bit cleaner 🙈

jeremiahlukus: done

PR #443, opened 2021-02-26
Migrate to null safety

miDeb:
Hi! I noticed that all dependencies of this package are ready for null safety, so I figured I could quickly migrate chewie to null safety.

I am aware that this is a fairly big change, but I hope that most of it is straightforward.
You'll notice that I have removed some null checks, simply because the checked value could never be null and the check would therefore always return `true`. For example, `_latestValue` (in Material-/CupertinoControls) is set right away in methods called from `initState`, so with null safety it can be marked as `late` and other methods don't have to do unnecessary null checks.

I tested various features in the example, and everything seems to work. I didn't migrate the example itself yet though, because `auto_orientation` does not yet support null safety ([I opened a pull request to resolve this](https://github.com/bytepark/auto_orientation/pull/16)).

There seems to be one test in the example, which fails for me _before and after_ this patch. Do you rely on manual testing only?


nstrelow:
BAAM, thanks a lot @miDeb .
I actually planned to look into null safety today. Pretty amazing that I now find a PR for it.
Will check you PR thoroughly and report back.

About `auto_orientation`, I wanted to check if it is actually still needed. It has not been updated since Dec 2019, so maybe flutter now supports its functionality.

miDeb:
> About auto_orientation, I wanted to check if it is actually still needed. It has not been updated since Dec 2019, so maybe flutter now supports its functionality.

Looks like you're right, in fact after simply removing the dependency rotating still seems to still work.

lib/src/cupertino_controls.dart
```diff
@@ -29,28 +29,33 @@ class CupertinoControls extends StatefulWidget {
...
-  Timer _initTimer;
+  Timer? _expandCollapseTimer;
+  Timer? _initTimer;
   bool _dragging = false;
 
-  VideoPlayerController controller;
-  ChewieController chewieController;
-  AnimationController playPauseIconAnimationController;
+  late VideoPlayerController controller;
+  ChewieController get chewieController => _chewieController!;
```
nstrelow, line 42: Why can we not use a late ChewieController here?

miDeb:
I have now migrated the example as well; note that I removed `auto_rotate.dart` as it was not used and did not seem to show additional features.

nstrelow:
Very awesome !!!

lib/src/cupertino_controls.dart
miDeb: The problem is the logic in `didChangeDependencies`. This is the only place where `_chewieController` is used directly, and this is required because it tries to compare the old ontroller with the new controller. That would not work with `late` the very first time a controller is set, because there is no old controller and a `LateInitializationError` would be triggered.

The way I wrote this the rest of the widget (especially build methods) don't have to do a null check every time, but `didChangeDependecies` can have access to a nullable `_chewieController`.

Hope that makes sense :)

example/lib/main.dart
```diff
@@ -37,17 +37,19 @@ class _ChewieDemoState extends State<ChewieDemo> {
   void dispose() {
     _videoPlayerController1.dispose();
     _videoPlayerController2.dispose();
-    _chewieController.dispose();
+    _chewieController?.dispose();
```
miDeb, line 40: This change is not strictly related to the migration, but it makes the test pass.

example/lib/main.dart
```diff
@@ -37,17 +37,19 @@ class _ChewieDemoState extends State<ChewieDemo> {
...
     _videoPlayerController1 = VideoPlayerController.network(
         'https://assets.mixkit.co/videos/preview/mixkit-forest-stream-in-the-sunlight-529-large.mp4');
-    await _videoPlayerController1.initialize();
     _videoPlayerController2 = VideoPlayerController.network(
         'https://assets.mixkit.co/videos/preview/mixkit-a-girl-blowing-a-bubble-gum-at-an-amusement-park-1226-large.mp4');
-    await _videoPlayerController2.initialize();
+    await Future.wait([
+      _videoPlayerController1.initialize(),
+      _videoPlayerController2.initialize()
+    ]);
```
miDeb, line 52: The changes here are also to make the test pass again.

lib/src/cupertino_controls.dart
nstrelow: Nice solution then.
Last idea I had was using initState, but then we cannot get the correct context for inheritedWidget.

Could you add a comment for the getter: Something like `// We know that _chewieController is set in didChangeDependencies`

miDeb: Sure, that will make it clearer for readers. Done.

nstrelow:
Awesome PR!
Thanks for bringing null safety to chewie.

Hope this will make people happy :D

nstrelow:
Published as 1.0.0

PR #444, opened 2021-02-27
Fix #438

### LINKED ISSUE #438: Refactor duration null check for _buildHitArea
Refactor material and cupertino controls to use the same duration checks:

https://github.com/brianegan/chewie/blob/ab14af6214e79de0b4d0ef65c5a151367035e1d7/lib/src/material_controls.dart#L64

https://github.com/brianegan/chewie/blob/ab14af6214e79de0b4d0ef65c5a151367035e1d7/lib/src/cupertino_controls.dart#L226

Material controls checks duration before rendering _buildHitArea, cupertino controls check it inside the _buildHitArea function.

gaganyadav80:


nstrelow:
This is probably not needed anymore or a lot changed because of the migration to null safety in #443 .

Thanks a lot anyway!

Also reviewing your changes: You seem to have added both null checks to both controllers. I meant either having the null check in buildHitArea or around buildHitArea.

PR #447, opened 2021-02-27
Animate the play/pause button in the bar as well

### LINKED ISSUE #446: The play/pause button in the material controls bar should animate as well
Just opening this issue for reference, I have a PR almost ready :)

Currently, the play/pause button in the center of the video that appears when pausing the video animates when toggling play/pause.
The button in the bar however does not animate.

miDeb:
Closes #446 

I moved the logic for forwarding/reversing the animation controller to `_updateState()` to catch all changes to `controller.value.isPlaying` regardless of how they were triggered.

I haven't made corresponding changes to the Cupertino controls, should I do that as well?

### DIFF
--- a/lib/src/material_controls.dart
+++ b/lib/src/material_controls.dart
@@ -36,6 +36,7 @@ class _MaterialControlsState extends State<MaterialControls>
   ChewieController get chewieController => _chewieController!;
   late AnimationController playPauseIconAnimationController =
       AnimationController(
+    value: controller.value.isPlaying ? 0 : 1,
     vsync: this,
     duration: const Duration(milliseconds: 400),
     reverseDuration: const Duration(milliseconds: 400),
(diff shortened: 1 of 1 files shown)


miDeb:
Here's a screen recording:


https://user-images.githubusercontent.com/44171190/109390369-86ecc180-7911-11eb-835a-027d58381410.mp4

nstrelow:
Nice :D
thanks a lot for the video as well.

We can merge that now, but I would prefer also having the code change in cupertino controls as well.

I wanted to look into merging as much as possible from cupertino and material controls. Ideally so that such changes are easily made for both without code duplication.

miDeb:
> I wanted to look into merging as much as possible from cupertino and material controls. Ideally so that such changes are easily made for both without code duplication.

Agree. There seems to be a lot of code duplication going on, with some features being only available in one implementation. If you want I can take a look at merging the implementations :)

miDeb:
Superseeded by #450, which covers all uses of play/pause buttons.

PR #449, opened 2021-02-28
Merge the Material and Cupertino progress bars

miDeb:
Still work in progress, so far I only merged the progress bars: The existing API still exists, but there is now a `VideoProgressBar` widget under the hood that takes a `ChewieProgressBarConfiguration` (for now).

### DIFF
--- a/lib/src/progress_bar.dart
+++ b/lib/src/progress_bar.dart
@@ -3,36 +3,29 @@ import 'package:flutter/material.dart';
 import 'package:flutter/widgets.dart';
 import 'package:video_player/video_player.dart';
 
-class ChewieProgressBarConfiguration {
-  const ChewieProgressBarConfiguration({
-    required this.barHeight,
-    required this.handleHeight,
-    required this.drawShadow,
-  });
-
-  final double barHeight;
-  final double handleHeight;
-  final bool drawShadow;
-}
-
 class VideoProgressBar extends StatefulWidget {
   VideoProgressBar(
     this.controller, {
-    required this.configuration,
     ChewieProgressColors? colors,
     this.onDragEnd,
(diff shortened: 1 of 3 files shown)


miDeb:
After thinking about this a bit more I have removed `ChewieProgressBarConfiguration`. I felt like introducing an additional object here is useless and confusing, so I made `barHeight`, `handleHeight` and `drawShadow` (these were previously the only members of `ChewieProgressBarConfiguration`) simply be constructor parameters.

nstrelow:
Very amazing, this gets rid of so much code already 🤩 

I will review it soon. But actually I think this would make a good fist step PR.

E.g. first move progressbar and then the others.

miDeb:
@nstrelow that would make sense as well. I made this ready for review.

miDeb:
@nstrelow that would make sense as well. I made this ready for review.

lib/src/progress_bar.dart
```diff
@@ -0,0 +1,206 @@
...
+class VideoProgressBar extends StatefulWidget {
+  VideoProgressBar(
+    this.controller, {
+    ChewieProgressColors? colors,
+    this.onDragEnd,
+    this.onDragStart,
+    this.onDragUpdate,
+    Key? key,
+    required this.barHeight,
+    required this.handleHeight,
```
miDeb, line 15: I only realized afterwards that `handleHeight` is `6` for both Material and Cupertino, but I then decided to leave it in.

nstrelow: That's fine!

nstrelow:
In german I would say: Wunschlos Glücklich!

miDeb:
Danke!

PR #450, opened 2021-03-01
Animate all play/pause icons

miDeb:
This PR adds an `AnimatedPlayPause` widget that encapsulates the animation. It implicitly animates the icon when it is rebuilt with a different `playing` argument.
Both `MaterialControls` and `CupertinoControls` can now simply use this widget without needing an `animationConttroller` on their own.

Screen recording:
https://user-images.githubusercontent.com/44171190/109505643-17a3d880-7a9d-11eb-9227-2dee7cc0862a.mp4

### DIFF
--- a/lib/src/animated_play_pause.dart
+++ b/lib/src/animated_play_pause.dart
@@ -38,6 +38,12 @@ class AnimatedPlayPauseState extends State<AnimatedPlayPause>
   }
 
   @override
+  void dispose() {
+    super.dispose();
+    animationController.dispose();
+  }
+
+  @override
   Widget build(BuildContext context) {
     return Center(
       child: AnimatedIcon(
--- a/lib/src/animated_play_pause.dart
+++ b/lib/src/animated_play_pause.dart
@@ -39,8 +39,8 @@ class AnimatedPlayPauseState extends State<AnimatedPlayPause>
 
   @override
   void dispose() {
-    super.dispose();
     animationController.dispose();
+    super.dispose();
   }
 
   @override
--- a/lib/src/animated_play_pause.dart
+++ b/lib/src/animated_play_pause.dart
@@ -1,0 +1,51 @@
+import 'package:flutter/material.dart';
+
+/// A widget that animates implicitly between a play and a pause icon.
+class AnimatedPlayPause extends StatefulWidget {
+  const AnimatedPlayPause({
+    Key? key,
+    required this.playing,
+    this.size,
+    this.color,
+  }) : super(key: key);
+
+  final double? size;
+  final bool playing;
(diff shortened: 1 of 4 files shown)


lib/src/material_controls.dart
```diff
@@ -379,7 +373,6 @@ class _MaterialControlsState extends State<MaterialControls>
 
     setState(() {
       if (controller.value.isPlaying) {
-        playPauseIconAnimationController.reverse();
```
miDeb, line 382: Fun fact: While `CupertinoControls` had an animated icon, it never called these methods on its controller, therefore never showing an animatoin.

nstrelow: Haha, great that it's working now :D

nstrelow:
Love the modularity.

Do you think it's alright that the icon animates now for iOS?


I just noticed, Only on flutter web, the play button seems uncentered somehow for material and cupertino. But this has been the case before this PR, so I'll merge this.

image

miDeb:
Let me investigate what's going on on the web. Also, looks like I forgot to dispose the animation controller...

miDeb:
Ok that took me a while... The bug on the web is caused by a different `visualDensity` on that platform (would have also affected desktop platforms). Finally found the solution here: https://github.com/flutter/flutter/issues/52980#issuecomment-709662594. I notices that adding the fix required doing it twice for material and cupertino, so I refactored that part out into a `CenterPlayButton` widget that can be shared.

miDeb:
> Do you think it's alright that the icon animates now for iOS?

From a UI perspective? I don't know. I mean, I don't even own an iPhone so I have no idea what it is supposed to look like on that platform or if it will look out of place.

lib/src/animated_play_pause.dart
```diff
@@ -37,6 +37,12 @@ class AnimatedPlayPauseState extends State<AnimatedPlayPause>
     }
   }
 
+  @override
+  void dispose() {
+    super.dispose();
```
nstrelow, line 42: super.dispose should be called last.
Cannot find the best practices for it, but disposing super first could kill everything before we dispose our stuff (at least that's what I am thinking)

nstrelow:
Oh wow, how annoying. Thanks a lot for getting to the bottom of this.

lib/src/animated_play_pause.dart
nstrelow: I changed this already and will merge now.

miDeb: Thanks!

jeremiahlukus:
Sweet i was looking at the code to see why the icon was jacked on flutter web. Thanks for fixing this

Issue #479, opened 2021-05-31
support custom RichText subtitles


ekuleshov:
Current subtitles support is limited to plain-text subtitles. Though some subtitle formats allow more advanced subtitle formatting. It would be great to support that in the custom subtitle builder.

Ahmadre:
No it's not limited to anything. Please read our Readme-section about the builder-method for your subtitles: https://github.com/brianegan/chewie#subtitles

ekuleshov:
> No it's not limited to anything. Please read our Readme-section about the builder-method for your subtitles

Please review my PR.

Yes, the subtitles builder can build custom rendering for subtitles, but the Subtitle class has the `text` field of type `String`. So, you can't stick custom data in there (and parsing it at the time when subtitles are shown is not a good idea. So, the main thing I changed type to `dynamic` and added example how to render RichText data

ekuleshov:
It seems like to make subtitles to use generic value instead of String will require generic type all the way up to the `Chewie` widget.

All in all, it seems like coupling subtitles into every video control implementation is too restrictive. It would be better to decouple subtitles from that and use a separate widget overlay instead. Somewhat similar to what `subtitle_wrapper_package` does. Then generic type only need to be specified when configuring subtitles overlay.

PR #480, opened 2021-05-31
support for custom text property for a subtitle

ekuleshov:
Support for custom RichText subtitles https://github.com/brianegan/chewie/issues/479


lib/src/chewie_player.dart
```diff
@@ -277,7 +277,7 @@ class ChewieController extends ChangeNotifier {
   final List<OptionItem> Function(BuildContext context)? additionalOptions;
 
   /// Define here your own Widget on how your n'th subtitle will look like
-  final Widget Function(BuildContext context, String subtitle)? subtitleBuilder;
+  final Widget Function(BuildContext context, dynamic subtitle)? subtitleBuilder;
```
Ahmadre, line 280: `dynamic` is dangerous and should always be avoided (see Flutter Code-Guidelines). If we want to provide a custom `text` and it belongs to the UI, it should be a Widget here, so the User can decide what kind of Widget she/he wants to add here.

lib/src/cupertino/cupertino_controls.dart
```diff
@@ -170,13 +170,18 @@ class _CupertinoControlsState extends State<CupertinoControls>
...
             color: const Color(0x96000000),
             borderRadius: BorderRadius.circular(10.0)),
-        child: Text(
-          currentSubtitle.first!.text,
-          style: const TextStyle(
-            fontSize: 18,
-          ),
-          textAlign: TextAlign.center,
-        ),
+        child: currentSubtitle.first!.text is InlineSpan
```
Ahmadre, line 173: `is` and `as` should be avoided. Instead provide directly the text as a Widget and you don't need this.

lib/src/material/material_controls.dart
```diff
@@ -244,13 +244,18 @@ class _MaterialControlsState extends State<MaterialControls>
...
           borderRadius: BorderRadius.circular(10.0),
         ),
-        child: Text(
-          currentSubtitle.first!.text,
-          style: const TextStyle(
-            fontSize: 18,
-          ),
-          textAlign: TextAlign.center,
-        ),
+        child: currentSubtitle.first!.text is InlineSpan
```
Ahmadre, line 247: see comment in: lib/src/cupertino/cupertino_controls.dart

lib/src/material/material_desktop_controls.dart
```diff
@@ -230,13 +230,18 @@ class _MaterialDesktopControlsState extends State<MaterialDesktopControls>
...
           borderRadius: BorderRadius.circular(10.0),
         ),
-        child: Text(
-          currentSubtitle.first!.text,
-          style: const TextStyle(
-            fontSize: 18,
-          ),
-          textAlign: TextAlign.center,
-        ),
+        child: currentSubtitle.first!.text is InlineSpan
```
Ahmadre, line 233: see comment in: see comment in: lib/src/cupertino/cupertino_controls.dart

lib/src/models/subtitle_model.dart
```diff
@@ -42,7 +45,10 @@ class Subtitle {
   final int index;
   final Duration start;
   final Duration end;
-  final String text;
+
+  /// Subtitle text. A [String] or another type, e.g. [InlineSpan], [TextSpan], etc.
+  /// Custom types need to be handled by [ChewieController.subtitleBuilder].
+  final dynamic text;
```
Ahmadre, line 51: see comment in: lib/src/chewie_player.dart

Ahmadre:
Thank you @ekuleshov for your PR!

lib/src/chewie_player.dart
ekuleshov: @Ahmadre using a `Widget` inside of `Subtitle` data object is probably not a good idea, i.e. to keep UI rendering separate from the values.

How about using generic value in `Subtitle<T>` and `Subtitles<T>` classes?

Then `T` could be `String`, `InlineSpan`. That may require to move the `subtitleBuilder` function into `Subtitles` to tie it in to the same generic type. Also this approach won't allow a mix of different text types like in my original PR.

ekuleshov:
> Thank you @ekuleshov for your PR!

You may want to reopen https://github.com/brianegan/chewie/issues/479

ekuleshov:
@Ahmadre I tried to change `Subtitles` and `subtitleBuilder` to be parametrized types like:

```
class Subtitles<T> {
  Subtitles(this.subtitle, { this.subtitleBuilder });

  final List<Subtitle<T>?> subtitle;
  Widget Function(BuildContext context, T subtitle)? subtitleBuilder;
  ...
```

The `subtitle` and `subtitleBuilder` values need to be tied together because of a common generic type, in under `Subtitles` class (which is basically a Typle) stored in the `ChewieController` instance.

Unfortunately I could not make generic types work. Getting a type case issue at runtime even without `is` and `as`.

Now I'm not sure if it is the best way to approach this.

Or perhaps maybe the subtitleBuilder could be changed to take in a `Subtitle` instance instead of a `String`. Then I could add an extension to `Subtitle` class and will be able to do my own type casting in the subtitleBulder implementation.

What do you think?

Ahmadre:
> @Ahmadre I tried to change `Subtitles` and `subtitleBuilder` to be parametrized types like:
> 
> 
> 
> ```
> 
> class Subtitles<T> {
> 
>   Subtitles(this.subtitle, { this.subtitleBuilder });
> 
> 
> 
>   final List<Subtitle<T>?> subtitle;
> 
>   Widget Function(BuildContext context, T subtitle)? subtitleBuilder;
> 
>   ...
> 
> ```
> 
> 
> 
> The `subtitle` and `subtitleBuilder` values need to be tied together because of a common generic type, in under `Subtitles` class (which is basically a Typle) stored in the `ChewieController` instance.
> 
> 
> 
> Unfortunately I could not make generic types work. Getting a type case issue at runtime even without `is` and `as`.
> 
> 
> 
> Now I'm not sure if it is the best way to approach this.
> 
> 
> 
> Or perhaps maybe the subtitleBuilder could be changed to take in a `Subtitle` instance instead of a `String`. Then I could extend `Subtitle` class with my own fields and will be able to do my own type casting in the subtitleBulder implementation.
> 
> 
> 
> What do you think?

thank you for your suggestions. I will have a look at it the next week. I'm on vacation :D

Ahmadre:
LGTM @ekuleshov :) Could you please update your branch with the latest master? After that I would merge this! I really like it and I think this way we got much more flexibility! 

And could you check the CI-build errors and correct them in your branch?

ekuleshov:
> LGTM

Which part? Like I said, I could not make generified variant work.

A variant with extending `Subtitle` turned out rather ugly. Personally I'd have preferred a variant with `dynamic` text.

ekuleshov:
It seems like to make subtitles to use generic value instead of String will require generic type all the way up to the `Chewie` widget.

All in all, it seems like coupling subtitles into every video control implementation is too restrictive. It would be better to decouple subtitles from that and use a separate widget overlay instead. Somewhat similar to what `subtitle_wrapper_package` does. Then generic type only need to be specified when configuring subtitles overlay.

Ahmadre:
I meant the actual changed with dynamic looks good!

We can do that like in the first step.

ekuleshov:
> We can do that like in the first step.

@Ahmadre okay. I re-did the variant with `dynamic` type. Please review

ekuleshov:
@Ahmadre do you have a timeline on getting this change into a plugin release?

PR #492, opened 2021-06-08
add copyWith extension | add resolution dialog

merdweebner:
lib/src/models/video_player_controller_extension.dart

### DIFF
--- a/example/lib/app/app.dart
+++ b/example/lib/app/app.dart
@@ -41,7 +41,7 @@ class _ChewieDemoState extends State<ChewieDemo> {
 
   Future<void> initializePlayer() async {
     _videoPlayerController1 = VideoPlayerController.network(
-        'https://assets.mixkit.co/videos/preview/mixkit-daytime-city-traffic-aerial-view-56-large.mp4');
+        'https://file-examples-com.github.io/uploads/2017/04/file_example_MP4_480_1_5MG.mp4');
     _videoPlayerController2 = VideoPlayerController.network(
         'https://assets.mixkit.co/videos/preview/mixkit-a-girl-blowing-a-bubble-gum-at-an-amusement-park-1226-large.mp4');
     await Future.wait([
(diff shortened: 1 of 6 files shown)


lib/src/material/material_controls.dart
```diff
@@ -179,6 +181,21 @@ class _MaterialControlsState extends State<MaterialControls>
       );
     }
 
+    if (chewieController.resolutions != null &&
```
diegotori, line 184: Replace with `if (chewieController.resolutions?.isNotEmpty ?? false) {`.

lib/src/material/material_controls.dart
```diff
@@ -442,6 +459,37 @@ class _MaterialControlsState extends State<MaterialControls>
     });
   }
 
+  Future<void> _onResolutionTap() async {
+    _hideTimer?.cancel();
+
+    final choosenResolution = await showModalBottomSheet<String>(
```
diegotori, line 465: Rename variable to `chosenResolution`.

example/lib/app/app.dart
```diff
@@ -52,6 +52,16 @@ class _ChewieDemoState extends State<ChewieDemo> {
       videoPlayerController: _videoPlayerController1,
```
maherjaafar, line 52: Same code is used between line 148 and 183. You could use the same _chewieController.

lib/src/material/material_controls.dart
Ahmadre: thx will do it

lib/src/material/material_controls.dart
Ahmadre: thx will do it

AhmadFalahian:
Can I use this branch without using master branch?

anwar907:
Hi, how about this future? @diegotori 
it's really helpful because `video_player` doesn't support different resolutions

diegotori:
@merdweebner are you still actively working on this PR?

dejancencelj:
Any update on casting?

PR #503, opened 2021-07-08
Add useRootNavigator property to ChewieController

ShakyaCsun:
The package always uses rootNavigator for Pushing Fullscreen Page, showing controls popup or bottom sheets, etc. But, there are cases where this behavior is undesirable and cause issues such as popping the wrong page on pressing back button (Especially when using custom navigation/routing packages such as [auto_route](https://pub.dev/packages/auto_route) or nested navigators).

I have added a useRootNavigator property to ChewieController such that the default behavior is same as before but users can configure chewie to not use rootNavigator.

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -70,7 +70,10 @@ class ChewieState extends State<Chewie> {
       _isFullScreen = true;
       await _pushFullScreenWidget(context);
     } else if (_isFullScreen) {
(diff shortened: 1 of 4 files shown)


diegotori:
@Ahmadre LGTM. Hopefully this gets added into the next release.

diegotori:
@Ahmadre gentle reminder regarding this PR. Thanks in advance.

PR #535, opened 2021-09-07
Update pubspec.yaml

tahaafg20:
Many packages are now using Provider 6.0.0 which creates a conflict with chewie which is using provider 5.0.0 in my project, checked and tested using provider 6.0.0 requires no changes and works just fine.
Please approve the change.
Thanks

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -11,7 +11,7 @@ dependencies:
   cupertino_icons: ^1.0.3
   flutter:
     sdk: flutter
-  provider: ^5.0.0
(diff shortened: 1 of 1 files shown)


diegotori:
@Ahmadre can we please get a fast-track approval on this one? Thanks in advance.

diegotori:
@Ahmadre a gentle reminder regarding this change. Thanks in advance.

tahaafg20:
Still having issues with this, provider 5.0.0 is really quite a few versions behind which is not supporting null safety either.

Issue #266, opened 2020-02-04
[notwork in fullscreen] auto play next src


shiyiya:
```dart
initPlayer() async {
   
    _videoPlayerController = VideoPlayerController.network(src);
    _videoPlayerController.addListener(autoNextLis);

    _chewieController = ChewieController(
      videoPlayerController: _videoPlayerController,
      aspectRatio: 16 / 9,
      autoPlay: true,
      looping: false,
      allowedScreenSleep: false,
      allowFullScreen: true,
      allowMuting: false,
    );
    setState(() {});
  }

 autoNextLis() {
    int total = _videoPlayerController.value.duration?.inMilliseconds;
    final int pos = _videoPlayerController.value.position?.inMilliseconds ?? 0;

    if (total == null) total = 1;
    if (total - pos <= 0) {
      _videoPlayerController.removeListener(autoNextLis);
      if (currPlayIndex + 1 < videoList.length) toggleVideo(currPlayIndex + 1);
    }
  }

 toggleVideo(int i) async {
    if (i == currPlayIndex ||
        i > videoList.length - 1 ||
        (!_videoPlayerController.value.initialized &&
            !_videoPlayerController.value.hasError)) return;
    await _videoPlayerController.pause();
    currPlayIndex = i;
    await initPlayer();
  }
```

When the playback is finished, the next video is played, and the full-screen video is stopped, but the non-full-screen video is playing the next video (click back to see)！

peakerWu:
@shiyiya   你好，全屏播放的时候，上一个播放完，自动播放下一个，你实现了吗

shiyiya:
> @shiyiya 你好，全屏播放的时候，上一个播放完，自动播放下一个，你实现了吗

你可以看我 fork 的那份 找跳转全屏（全屏播放）那个代码文件的提交记录，原项目的判断似乎有点问题 我改了一下就好了

peakerWu:
> > @shiyiya 你好，全屏播放的时候，上一个播放完，自动播放下一个，你实现了吗
> 
> 你可以看我 fork 的那份 找跳转全屏（全屏播放）那个代码文件的提交记录，原项目的判断似乎有点问题 我改了一下就好了
好的 非常谢谢  我去看下~

peakerWu:
> > @shiyiya 你好，全屏播放的时候，上一个播放完，自动播放下一个，你实现了吗
> 
[...]

PR #547, opened 2021-09-19
Fix full screen listener #266

shiyiya:
Hold fullscreen state when ctrl changes

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -42,6 +42,8 @@ class Chewie extends StatefulWidget {
 
 class ChewieState extends State<Chewie> {
   bool _isFullScreen = false;
+
+  bool get isFullScreen => widget.controller.isFullScreen;
   late PlayerNotifier notifier;
 
   @override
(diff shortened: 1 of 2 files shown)


lib/src/chewie_player.dart
```diff
@@ -42,6 +42,8 @@ class Chewie extends StatefulWidget {
 
 class ChewieState extends State<Chewie> {
   bool _isFullScreen = false;
+
+  bool get isFullScreen => widget.controller.isFullScreen;
```
diegotori, line 46: Might be better to rename this to `isControllerFullScreen` so that it's distinct from `_isFullScreen`.

diegotori (changes requested):

lib/src/chewie_player.dart
```diff
@@ -63,15 +65,20 @@ class ChewieState extends State<Chewie> {
...
     super.didUpdateWidget(oldWidget);
+    if (_isFullScreen != isControllerFullScreen) {
+      widget.controller._isFullScreen = _isFullScreen;
+    }
   }
 
   Future<void> listener() async {
-    if (widget.controller.isFullScreen && !_isFullScreen) {
-      _isFullScreen = true;
+    if (isControllerFullScreen && !_isFullScreen) {
```
Ahmadre, line 74: Totally unnecessary check here. Just do:

```dart
if (isControllerFullScreen != _isFullScreen)
...
else {...}
```

PR #562, opened 2021-10-18
Fix - upgrade dependencies

maherjaafar:
This PR contains:

- [x] Upgrade dependencies
- [x] Bump version to `1.2.3`


diegotori:
@Ahmadre gentle reminder that we could use a release as a result of this change. Thanks in advance.

PR #600, opened 2022-02-10
Create control to close the player and exit the current page like Instagram

AhmadFalahian:
There is no control in your player to close the video player. For example, suppose we are in full screen mode and we want to stop the whole player and exit it by pressing the close button, like the mode that Instagram has for its own player.


diegotori:
@AhmadFalahian Looks like this change needs to be scoped to just the changes to the player. Please rebase your changes from the latest master changes and re-submit this PR. Thank in advance.

example/lib/app/app.dart
```diff
@@ -133,6 +133,8 @@ class _ChewieDemoState extends State<ChewieDemo> {
               ),
       ),
 
+      hideControlsTimer: const Duration(seconds: 1),
```
maherjaafar, line 142: We can add a variable: 

```dart
({
... // Your variable
this.createControlsTimer = const Duration(seconds: 1),
)};
```

```dart
final Duration createControlsTimer;
```

It would be better if the developer is able to modify it.

Ortes:
@diegotori This PR has no actual diff to review — it compares `master` against the old `feature-add-new-cupertino-skip-buttons` branch (frozen since Feb 2022), which is why it shows 500+ unrelated commits. I checked AhmadFalahian's fork across every branch and found no commits authored by them anywhere — this was effectively a feature request opened as a PR, with no code ever attached.

The underlying ask (a quick way to exit fullscreen) is now covered by #930, already merged into `master`.

Suggest closing this one.

PR #616, opened 2022-03-07
ChewieControllerProvider doesn't provide ChewieController's fields

guillempuche:
# Context
In my understanding is that`MaterialControls` uses `ChewieController.of` because wants to access the nearest `ChewieControllerProvider` on the tree, but the parent will be a `ChewieControllerProvider` instead of `ChewieController`.

`ChewieControllerProvider` isn't used on other parts (check it [here](https://github.com/fluttercommunity/chewie/search?q=ChewieControllerProvider). 

# Problem
The private field `_chewieController` in class `MaterialControls` ([line 130](https://github.com/fluttercommunity/chewie/blob/b131d6938a7fc863895e94f4c0e83e96679eabed/lib/src/material/material_controls.dart#L130)) uses `ChewieControllerProvider`, but this doesn't have the initialized fields of `ChewieController` (e.g. `allowMuting`, etc).

Why? Imaging you have this code:
```
_videoChewieController = ChewieController(
        allowingMuting: false,
);
```
The muting button is shown on the controls, but it shouldn't show.

# Solution
I think few people use `ChewieControllerProvider`, for this reason I've changed the code in a way that the developer has to pass the type `ChewieControllerProvider` if he wants to use it.

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -495,11 +495,14 @@ class ChewieController extends ChangeNotifier {
   /// Defines a custom RoutePageBuilder for the fullscreen
   final ChewieRoutePageBuilder? routePageBuilder;
 
-  static ChewieController of(BuildContext context) {
+  static ChewieController of<ChewieControllerProvider>(BuildContext context) {
+    if (null is ChewieControllerProvider) {
+      return context.dependOnInheritedWidgetOfExactType<ChewieController>()!;
+    }
     final chewieControllerProvider =
         context.dependOnInheritedWidgetOfExactType<ChewieControllerProvider>()!;
 
-    return chewieControllerProvider.controller;
+    return chewieControllerProvider.controller; 
   }
 
   bool _isFullScreen = false;


diegotori:
@guillempuche looks like it failed both lint and formatting checks. Please address those when you're able to.

As for the proposed changes, I'm trying to understand the purpose of these changes based on your explanation. So what you're saying is that in `MaterialControls` (and to an extent, `CupertinoControls`), when `didUpdateDependencies()` is called, that `ChewieController.of` isn't returning the `ChewieController` and is instead returning the `ChewieControllerProvider`?

My understanding is that `ChewieController.of` always looks up the `ChewieControllerProvider` internally before returning its `ChewieController` instance. Every time you create a `ChewiePlayer`, it always provides the `ChewieController` above it via `ChewieControllerProvider`. Just confused as to what these changes bring to the table.

diegotori:
@brianegan I know that you're a busy man, but I would like to hear your thoughts on the above change, since you have more familiarity with the the decisions made when you initially wrote the library. Thanks in advance.

brianegan:
> In my understanding is thatMaterialControls uses ChewieController.of because wants to access the nearest ChewieControllerProvider on the tree, but the parent will be a ChewieControllerProvider instead of ChewieController.

Not quite! The `ChewieControllerProvider` is  just an `InheritedWidget` that passes the `ChewieController` down the Widget tree. 

Therefore, `MaterialControls` uses `ChewieController.of` because wants to access the nearest **`ChewieController`**, not the  the `ChewieControllerProvider`.

Overall, this code change does not make much sense to me. Could you please describe what you're trying to achieve?

lib/src/chewie_player.dart
```diff
@@ -495,11 +495,14 @@ class ChewieController extends ChangeNotifier {
   /// Defines a custom RoutePageBuilder for the fullscreen
   final ChewieRoutePageBuilder? routePageBuilder;
 
-  static ChewieController of(BuildContext context) {
+  static ChewieController of<ChewieControllerProvider>(BuildContext context) {
+    if (null is ChewieControllerProvider) {
+      return context.dependOnInheritedWidgetOfExactType<ChewieController>()!;
```
brianegan, line 500: `dependOnInheritedWidgetOfExactType` must depend on an `InheritedWidget`. `ChewieController` is not an `InheritedWidget`, it is a `ChangeNotifier`. This code will not work.

lib/src/chewie_player.dart
```diff
@@ -495,11 +495,14 @@ class ChewieController extends ChangeNotifier {
   /// Defines a custom RoutePageBuilder for the fullscreen
   final ChewieRoutePageBuilder? routePageBuilder;
 
-  static ChewieController of(BuildContext context) {
+  static ChewieController of<ChewieControllerProvider>(BuildContext context) {
```
brianegan, line 498: `ChewieControllerProvider` is not a Generic Class. It is a specific class used to pass the `ChewieController` to any descendant widget. It should not be used as a generic here.

diegotori:
@guillempuche are you still working on this PR? If so please read the feedback carefully so that you can determine whether to address these changes. Thanks in advance.

guillempuche:
Okay, long time I made a hack to solve this.

@diegotori, better to close the PR, for now.

If in the future this also happens to someone else, you'll have more info.

PR #617, opened 2022-03-08
Allow video zooming with InteractiveViewer widget.

jmsanc:
Hi,
This might be (or not!) an interesting feature for Chewie. It allows for zoom and panning in the video window. Please feel free to consider it as a potential feature. 
Zoom/panning and scale factor can be configured from the hosting app side.
Thank you.

### DIFF
--- a/lib/src/player_with_controls.dart
+++ b/lib/src/player_with_controls.dart
@@ -37,11 +37,16 @@ class PlayerWithControls extends StatelessWidget {
         children: <Widget>[
           if (chewieController.placeholder != null)
             chewieController.placeholder!,
-          Center(
-            child: AspectRatio(
-              aspectRatio: chewieController.aspectRatio ??
-                  chewieController.videoPlayerController.value.aspectRatio,
-              child: VideoPlayer(chewieController.videoPlayerController),
+          InteractiveViewer(
+            maxScale: chewieController.maxScale,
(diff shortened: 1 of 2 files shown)


lib/src/player_with_controls.dart
```diff
@@ -52,14 +57,17 @@ class PlayerWithControls extends StatelessWidget {
...
               ) =>
-                  AnimatedOpacity(
-                opacity: notifier.hideStuff ? 0.0 : 0.8,
-                duration: const Duration(
-                  milliseconds: 250,
-                ),
-                child: Container(
-                  decoration: const BoxDecoration(color: Colors.black54),
-                  child: Container(),
+                  Visibility(
```
diegotori, line 60: Just curious as to why the `Visibility` widget above the `AnimatedOpacity` one. Is it because the latter interferes with `InteractiveViewer`?

jmsanc: That's correct. Even with opacity at zero the gestures will not be picked by the InteractiveViewer. So if we hide the AnimatedOpacity it would work correctly.

diegotori:
LGTM. @maherjaafar please let me know if you concur. Thanks.

diegotori:
@maherjaafar last call for a review. Otherwise, I'm merging it first thing, then releasing a new build later that day. Thanks.

PR #505, opened 2021-07-13
Add YT suppport to chewie

harishsg99:
This support can help lots of developers to integrate Video player which can play video from youtube using chewie into their  flutter app

PR #619, opened 2022-03-16
Add ext_video_player for Youtube support

prateekmedia:
A successor of #505


prateekmedia:
And will the new implementation needed to be renamed? And if yes what will it be.

PR #621, opened 2022-04-01
Upgrade kotlin version and compileSdkVersion in example

Henri2h:
Hello !

Quick pull request to upgrade
* kotlin version
* android compileSdkVersion
in the example. (to make it compile :) )


diegotori:
@Henri2h LGTM. If it passes lint and format checks, then it'll get merged.

PR #669, opened 2022-09-07
Fix for CenterPlayButton UI bug when using Material 3

luis901101:
When using Material 3, the AnimatedOpacity of **CenterPlayButton** expands to fill the available space, and looks like this:

IMG_0638

This PR fixes it, and in addition:
- It exposes MaterialDesktopControls to allow devs to explicitly use it.
- Some lints fixed for better performance

### DIFF
--- a/lib/src/center_play_button.dart
+++ b/lib/src/center_play_button.dart
@@ -21,14 +21,14 @@ class CenterPlayButton extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return Container(
+    return ColoredBox(
       color: Colors.transparent,
       child: Center(
         child: UnconstrainedBox(
           child: AnimatedOpacity(
             opacity: show ? 1.0 : 0.0,
             duration: const Duration(milliseconds: 300),
-            child: Container(
+            child: DecoratedBox(
               decoration: BoxDecoration(
                 color: backgroundColor,
                 shape: BoxShape.circle,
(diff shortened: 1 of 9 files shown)


lib/src/center_play_button.dart
```diff
@@ -21,23 +21,23 @@ class CenterPlayButton extends StatelessWidget {
...
-              padding: const EdgeInsets.all(12.0),
+        child: UnconstrainedBox(
+          child: AnimatedOpacity(
+            opacity: show ? 1.0 : 0.0,
+            duration: const Duration(milliseconds: 300),
+            child: DecoratedBox(
+              decoration: BoxDecoration(
+                color: backgroundColor,
+                shape: BoxShape.circle,
+              ),
```
diegotori, line 35: @luis901101 Thanks for the PR. Quick question. Does this change in any way break non-Material 3 setups?

For example, if the user wants to not use Material 3, does this change create an impact?

If so, you may need to only apply this if the current `Theme` has Material 3 enabled, so that it doesn't break any users who don't want to opt-in to that at the moment.

diegotori (changes requested):

lib/src/center_play_button.dart
luis901101: Hi, I tested both cases, Material3 and Non-Material3 themes and it works as expected in both cases.

diegotori (approved):

diegotori:
Gonna try them out on my end very shortly, but if they look good, then they'll get merged.

diegotori:
LGTM, merging very shortly.

PR #675, opened 2022-09-30
Added 10 seconds forward and backward functionality in the controls

ppmurmu:
- changed "LIVE" text colour to red for better UI.
- 10 seconds forward and backward functionality is only visible when "isLive" property is set to false.
- Changed the container's colour to black with 0.35 opacity for better visibility of controls.


Ortes:
For triage: the ±10s controls this adds shipped in #703 (Sep 2024), so merging this now would render a second pair of seek buttons in the bottom bar, and the seek here isn't clamped to the video bounds. The unrelated tweaks (red LIVE label, bar overlay) would need their own PR against current master. Suggest closing as superseded by #703.

Issue #697, opened 2023-01-15
Fix chewie player example in chewie package on pub.dev


zyllus17:
Currently the example on chewie links to [example/lib/main.dart](https://github.com/fluttercommunity/chewie/blob/master/example/lib/main.dart) whereas it should link to [example/lib/app/app.dart](https://github.com/fluttercommunity/chewie/blob/master/example/lib/app/app.dart) because of this it shows this in example section of pub.dev

<img width="1426" alt="Screenshot 2023-01-15 at 5 49 01 PM" src="https://user-images.githubusercontent.com/50977126/212540179-9934584c-e955-4fda-b09e-b64fec10c292.png">

If possible please simplify the example app as well. If needed I can create a PR with a simple example as well.

PR #698, opened 2023-01-15
Updated chewie player for pub.dev

zyllus17:
Updated the main.dart file with a simpler example and null safety to start working with Chewie player. 

Reason for creating this example - Issue #697 

If approved this will fix the example mentioned in Issue #697 

image

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -1,8 +1,77 @@
+import 'package:chewie/chewie.dart';
 import 'package:chewie_example/app/app.dart';
 import 'package:flutter/material.dart';
+// ignore_for_file: depend_on_referenced_packages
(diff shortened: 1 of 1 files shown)


diegotori:
This is not gonna fly in its current form, simply because it will override the more comprehensive example that already exists, which should be retained.

If you want to keep both, you have to incorporate the existing demo app into this PR. Otherwise, we're essentially wiping out useful code in this PR.

zyllus17:
I can create a new file with a name `chewie_video_player.dart` and place the widget in there but then it won't fix the issue of showing the correct example in the `pub.dev`. Will it be possible to point to a different file, prefereably `app.dart` from your end so that the correct example can be shown on pub.dev

PR #703, opened 2023-01-28
Adding Seek buttons for Android

### LINKED ISSUE #615: [feature-request] For android forward and backward controls like iOS
Hello,
I am using the chewie package for android and ios for video playing. In iOS, we have forward and backward controls so anyone can forward 15 sec or backward the 15 sec but on android, it is not possible can we have a double tap on right to forward it and on the left side 15 backward.

Thank you

GyanendroKh:
Adding Forward and Backward seek buttons for Android.
Fixes #615

### DIFF
--- a/lib/src/center_seek_button.dart
+++ b/lib/src/center_seek_button.dart
@@ -7,6 +7,8 @@ class CenterSeekButton extends StatelessWidget {
     required this.backgroundColor,
     this.iconColor,
     required this.show,
+    this.fadeDuration = const Duration(milliseconds: 300),
+    this.iconSize = 26,
     this.onPressed,
   }) : super(key: key);
 
@@ -15,6 +17,8 @@ class CenterSeekButton extends StatelessWidget {
   final Color? iconColor;
   final bool show;
   final VoidCallback? onPressed;
+  final Duration fadeDuration;
+  final double iconSize;
 
   @override
   Widget build(BuildContext context) {
@@ -24,7 +28,7 @@ class CenterSeekButton extends StatelessWidget {
         child: UnconstrainedBox(
           child: AnimatedOpacity(
             opacity: show ? 1.0 : 0.0,
-            duration: const Duration(milliseconds: 300),
+            duration: fadeDuration,
             child: DecoratedBox(
               decoration: BoxDecoration(
                 color: backgroundColor,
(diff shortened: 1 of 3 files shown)


diegotori:
If you can accomplish this for `CupertinoControls`, then I would accept this PR.

@brianegan any thoughts on this proposed change, since that would be nice if users were able to seek.

lib/src/center_seek_button.dart
```diff
@@ -0,0 +1,47 @@
...
+            duration: const Duration(milliseconds: 300),
+            child: DecoratedBox(
+              decoration: BoxDecoration(
+                color: backgroundColor,
+                shape: BoxShape.circle,
+              ),
+              // Always set the iconSize on the IconButton, not on the Icon itself:
+              // https://github.com/flutter/flutter/issues/52980
+              child: IconButton(
+                iconSize: 26,
```
diegotori, line 36: This should be configurable from this `Widget`'s constructor, defaulting to `26`.

Then, there should be a property in `ChewieController` that configures it, and when creating this widget, you pass in the current value from the controller.

lib/src/center_seek_button.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return ColoredBox(
+      color: Colors.transparent,
+      child: Center(
+        child: UnconstrainedBox(
+          child: AnimatedOpacity(
+            opacity: show ? 1.0 : 0.0,
+            duration: const Duration(milliseconds: 300),
```
diegotori, line 27: This should also be configurable from this `Widget`'s constructor, defaulting to `const Duration(milliseconds: 300)`.

Then, there should be a property in `ChewieController` that configures it, and when creating this widget, you pass in the current value from the controller.

diegotori (changes requested):

lib/src/center_seek_button.dart
GyanendroKh: Added.

lib/src/center_seek_button.dart
GyanendroKh: Added.

andes2912:
Please merge this PR, because Andriod need this feature

Ahmadre:
> Please merge this PR, because Android need this feature

We can't merge changes without investigating and the changes are done which @diegotori requested. Please update your PR accordingly and we will test this :)

diegotori:
@Ahmadre if you have capacity, please review his changes by running them locally.

I wanna say that there were exceptions thrown due to the current library calling `setState` while in the middle of one.

Thanks.

Ahmadre:
LGTM @diegotori 

Your change requests seems to be implemented. We merge this if you want.

diegotori (approved):
LGTM.

diegotori:
@Ahmadre can you please submit a follow-up PR that fixes lint issues related to this one? Thanks.

Issue #665, opened 2022-08-29
controller copyWith not effective in fullscreen


vincepunkrock:
I am changing the overlay when the video is playing. I added a listener to my controller and in it, I check the isPlaying property. If the video is playing, I change the overlay by updating the controller using the copyWith method (in a setState). 
It works well when the video is not in fullscreen. 

When in fullscreen, the overlay doesn't get changed, although isPlaying is true and the line where I update the controller gets executed.

diegotori:
Can you please submit a example project demonstrating this issue? That way I can better diagnose it. Thanks.

vincepunkrock:
This is my widget:

import 'package:cached_network_image/cached_network_image.dart';
...
import 'package:chewie/chewie.dart';
import 'package:flutter/material.dart';
import 'package:flutter/services.dart';
import 'package:video_player/video_player.dart';

class MyVideoPlayer extends StatefulWidget {
  final Video video;
  final double height;
  final double width;
  MyVideoPlayer({required this.video, required this.height, required this.width});

  @override
  _MyVideoPlayerState createState() => _MyVideoPlayerState();
}

class _MyVideoPlayerState extends State<MyVideoPlayer> {
  VideoPlayerController? _controller;
  ChewieController? _chewieController;
  bool _overlayRemoved = false;

  @override
  void initState() {
    super.initState();
    initializePlayer();
  }

  Future<void> initializePlayer() async {

    _controller = VideoPlayerController.network(widget.video.mp4);
    await Future.wait([
      _controller!.initialize()
    ]);
    _chewieController = ChewieController(
      videoPlayerController: _controller!,
      deviceOrientationsAfterFullScreen: [
        DeviceOrientation.portraitDown,
        DeviceOrientation.portraitUp
      ],
      deviceOrientationsOnEnterFullScreen: [
        DeviceOrientation.landscapeLeft,
        DeviceOrientation.landscapeRight
      ],
      autoPlay: false,
      looping: false,
      allowPlaybackSpeedChanging: false,
      allowMuting: false,
      autoInitialize: false,
      overlay: Center(child: CachedNetworkImage(imageUrl: widget.video.thumbnail)),
      placeholder: Center(child: CachedNetworkImage(imageUrl: widget.video.thumbnail, )),
      customControls: Theme.of(context).platform == TargetPlatform.iOS ? MyCupertinoControls(
        backgroundColor: const Color.fromRGBO(41, 41, 41, 0.7),
[...]

PR #713, opened 2023-03-04
Update for the overlay builder

murphycrosby:
#665

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -257,6 +257,7 @@ class ChewieController extends ChangeNotifier {
     this.materialProgressColors,
     this.placeholder,
     this.overlay,
+    this.overlayBuilder,
     this.showControlsOnInitialize = true,
     this.showOptions = true,
     this.optionsBuilder,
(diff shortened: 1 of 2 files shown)


lib/src/chewie_player.dart
```diff
@@ -390,6 +393,9 @@ class ChewieController extends ChangeNotifier {
 
   static const defaultHideControlsTimer = Duration(seconds: 3);
 
+  /// Define here your own Widget on how your overlay will look
+  Widget Function(BuildContext context)? overlayBuilder;
```
diegotori, line 397: This should be `final` since we're trying to make the controller immutable.

lib/src/player_with_controls.dart
```diff
@@ -20,6 +20,16 @@ class PlayerWithControls extends StatelessWidget {
...
     }
 
+    Widget buildOverlay(BuildContext context) {
+      if(chewieController.overlayBuilder != null) {
+        return chewieController.overlayBuilder!(context);
+      }
+      if (chewieController.overlay != null) {
+        return chewieController.overlay!;
+      }
+      return const SizedBox.shrink();
```
diegotori, line 30: This should be chained as an if-else statement, with it defaulting to `SizedBox.shrink()`.

For bonus credit, instead of using a function to compute the correct Widget to return, why not move it into a standalone widget that takes in the current controller and computes it when `build` is called on it.

diegotori (changes requested):

diegotori:
@murphycrosby if you have the time to work on the requested changes, then I'll be able to review them.

Thanks in advance.

PR #740, opened 2023-06-18
Add parameter to allow user to set bottom bar backgrond color along w…

krishnajithapa:
There was an issue reported where user was not able to see bottom bar button when background color was white. So i have added paremeter to let user specify the background and button color.

### DIFF
--- a/example/lib/app/app.dart
+++ b/example/lib/app/app.dart
@@ -188,10 +188,12 @@ class _ChewieDemoState extends State<ChewieDemo> {
                             .videoPlayerController.value.isInitialized
                     ? Chewie(
                         controller: _chewieController!,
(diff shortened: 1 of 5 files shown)


diegotori:
@krishnajithapa thanks for your PR.

I think that the changes should be present in `ChewieController` if they are not overridden in `Chewie` by the user.

In other words, `ChewieController` should hold immutable `Color` properties related to this PR, and they should be used in `MaterialControls` as well as `MaterialDesktopControls`. If those two classes specify an optional custom color value, then use those values, otherwise fallback to the controller.

`Chewie`, `PlayerWithControls`, and `AdaptiveControls` should funnel the user-defined optional `Color` properties into `MaterialControls` and MaterialDesktopControls`.

PR #753, opened 2023-07-06
Update Dart SDK to version 3

yangsfang:
The chewie package only builds successfully with Dart SDK 3.0. This commit updates the documentation to be reflected on pub.dev. This commit does not change the fact that the package fails to build under Dart SDK 2.17 - 2.19.

Issue https://github.com/fluttercommunity/chewie/issues/752

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -4,8 +4,8 @@ version: 1.6.0+1
 homepage: https://github.com/fluttercommunity/chewie
 
 environment:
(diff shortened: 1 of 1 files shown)


diegotori:
@yangsfang Hold off on this PR, since I'm in the process of addressing it in `wakelock_plus`.

diegotori:
I just published version `1.7.0` which should now be directly compatible with at least Dart `2.18` and Flutter `3.3.0`.

PR #769, opened 2023-08-23
Adding Watermark Feature: Allow Adding Watermark Widgets to Videos

amrhassann:
Changes Made:

Added a new watermark feature that allows users to add watermark widgets to videos.
Extended the existing classes to accommodate watermark rendering and positioning.

How to Use:
simply any one can add a watermark like that:

_chewieController!.watermark = const Center(child: Text("Hello Watermark Widget",style: TextStyle(color: Colors.white),));

Benefits:
Sometimes the requirements of the application require some security for the video and knowing a specific sign indicating the source of the video leak
It is just to increase safety

Thanks for your time and support
Best regards,
Amr Hassan

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -246,6 +246,7 @@ class ChewieState extends State<Chewie> {
 class ChewieController extends ChangeNotifier {
   ChewieController({
     required this.videoPlayerController,
+    this.watermark,
     this.optionsTranslation,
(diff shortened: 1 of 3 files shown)


diegotori:
Thank you for your PR, @amrhassann.

Quick question, is there any reason why you can't use `ChewieController.overlay` to achieve the same thing?

That's why it's there, so that your watermark can sit between the video and the controls.

PR #777, opened 2023-09-13
fix display size while Chewie wrapped by some rotate widget

bailyzheng:
When Chewie wrapped by RotatedBox like this. It is used in the situation that device not support landscape and portrait switch.
```
RotatedBox(quarterTurns: 1,
                      child: Chewie(
                        controller: _chewieController!,
                      ))
```
It will not occupy all the space.
S1

Use LayoutBuilder instead of MediaQuery is more adaptive. It looks like this now.
s2

### DIFF
--- a/lib/src/player_with_controls.dart
+++ b/lib/src/player_with_controls.dart
@@ -83,15 +83,18 @@ class PlayerWithControls extends StatelessWidget {
       );
     }
 
-    return Center(
-      child: SizedBox(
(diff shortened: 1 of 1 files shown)


diegotori:
LGTM. Will try out your changes either tonight or tomorrow night.

diegotori:
@bailyzheng sorry for the delay. Will get a chance to run it on my end first thing tomorrow.

PR #780, opened 2023-10-04
Feat  add the ability to style bottomsheet items

DavidCarrilho:


diegotori:
@DavidCarrilho just trying to understand what value this adds to the library?

In other words, what is this PR trying to achieve that cannot be done currently with the library?

DavidCarrilho:
Hello @diegotori ,

Thank you for taking the time to review my PR. I'd be happy to explain the value it brings to the library.

The main motivation behind this PR is to enhance the visual adaptability of the bottomsheet items. As we know, depending on the theme or the specifics of the bottomsheet in use, the default styling might not always provide optimal visibility or aesthetics. This can lead to scenarios where the items on the bottomsheet might be hard to read, visually jarring, or simply not in line with the desired look and feel of an app.

While the current library is robust and offers great functionality, it seems to lack flexibility in terms of styling these bottomsheet items. My PR introduces this capability, allowing developers to adjust styles as per their requirements or the design language of their application. This means that with these changes, users of the library can ensure that their bottomsheets always look great and are consistent with the rest of their UI, irrespective of the theme they are using.

I believe this addition would make the library more versatile and user-friendly, catering to a broader range of design needs and ensuring a better user experience.

I hope this clarifies the intent behind my contribution. Please let me know if there are any other questions or if further details are needed.

PR #784, opened 2023-10-10
This will allow us to handle the video rotation, as currently getting flipped videos

fisforfaheem:


diegotori:
@fisforfaheem I need more context in terms of this change.

In other words, what is this attempting to solve versus what's already available? Chewie supports setting one or more system orientations that can be applied to the video player and when you exit out of it.

fisforfaheem:
I was working on a project and the video I was getting as output in the player was flipped .. If I record it in flipped...horizontal states etc... adding this..( after some improvements may help other as well) Onemob mobile app is one example

diegotori:
@fisforfaheem if this is happening on Android, then please raise an issue with the Flutter team related to `video_player`.

Keep in mind that this library is merely a UI wrapper around `VideoPlayer`. If the issue is happening there, then it's on the Flutter team to address so that we don't prematurely attempt to rotate the video if there's no need to do so.

I'm not a big fan of trying to apply another `Transform` on top of the one that `VideoPlayer` applies based on the rotation correction values that it receives from the platform when playing back a video.

fisforfaheem:
Sure got your point

diegotori:
As of version `1.7.3`, which I just published today, this PR may not be needed anymore.

In other words, we now use a `LayoutBuilder` to build Chewie's player. As a result, you can rotate it like so:

```dart
RotatedBox(
  quarterTurns: 1,
  child: Chewie(
    controller: _chewieController!,
  ),
)
```

diegotori:
I'm going to close this PR since this is no longer needed due to it being addressed in version `1.7.5`.

fisforfaheem:
Ok

PR #794, opened 2023-11-05
Manually select desired controls

vinaykumarrock986612:
Currently, the controls are displayed based on the platform, and there's an option to choose build custom controls. But there's no option to use Cupertino controls as default if I want.

Added a high-level attribute to select desired controls type

### DIFF
--- a/lib/src/helpers/adaptive_controls.dart
+++ b/lib/src/helpers/adaptive_controls.dart
@@ -1,13 +1,32 @@
 import 'package:chewie/chewie.dart';
 import 'package:flutter/material.dart';
 
+enum ControlsType { cupertino, material, materialDesktop, adaptive }
+
 class AdaptiveControls extends StatelessWidget {
   const AdaptiveControls({
     Key? key,
+    required this.controlsType,
   }) : super(key: key);
+
+  final ControlsType controlsType;
 
   @override
   Widget build(BuildContext context) {
+    switch (controlsType) {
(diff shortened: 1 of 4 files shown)


lib/src/helpers/adaptive_controls.dart
```diff
@@ -1,13 +1,32 @@
...
+          iconColor: Color.fromARGB(255, 200, 200, 200),
+        );
+      case ControlsType.material:
+        return const MaterialControls();
+      case ControlsType.materialDesktop:
+        return const MaterialDesktopControls();
+      case ControlsType.adaptive:
+        break;
+    }
+
```
diegotori, line 29: Instead of having to duplicate code in this method, consider doing the following:

* Instead of using a separate enum for this, you may want to use an optional and nullable `TargetPlatform` when creating this widget. That way if it exists, use that value, otherwise, fallback to the one from the theme.
* Then you can determine the effective `TargetPlatform` to use based on the above and pass that into the switch statement:

```dart
final effectiveTargetPlatform = customPlatform ?? Theme.of(context).platform;
switch (effectiveTargetPlatform) {
  //...
}
```

diegotori (changes requested):

lib/src/helpers/adaptive_controls.dart
maherjaafar: @vinaykumarrock986612 any updates?

Issue #789, opened 2023-10-20
Progress bar does not follow drag


koutaro-masaki:
When you grab and slide the progress bar, you would expect the pointer to follow your finger, but Chewie doesn't do that.
It seems that the `_latestDraggableOffset` is not passed to the `StaticProgressBar` and the update is not reflected, is this intended?

https://github.com/fluttercommunity/chewie/pull/712

basemkhirat:
good to fix this

ashanftw:
https://github.com/fluttercommunity/chewie/issues/797

im having the same issue

diegotori:
@koutaro-masaki I'll look into this sometime this week. If you have a PR that fixes this, I'll be happy to review it. Thanks.

ashanftw:
@diegotori Would you be able to publish the fix to the package?

diegotori:
I'll make an effort first thing tomorrow. Gonna be out of the country for two weeks, so I'll try to squeeze it before I head out.

diegotori:
Published in version `1.7.2`.

ashanftw:
Thank you!

PR #798, opened 2023-11-27
Fix: Progress bar does not follow drag #789

koutaro-masaki:
This PR fixes issue #789 and the following bugs that were uncovered as a result of fixing it:
- Due to using Duration.zero as an invalid value, the pointer's behavior at the left edge of the progress bar was unnatural. 
- The pointer was able to go beyond the confines of the progress bar.

### DIFF
--- a/lib/src/progress_bar.dart
+++ b/lib/src/progress_bar.dart
@@ -266,7 +266,7 @@ extension RelativePositionExtensions on BuildContext {
     if (globalPosition == null) return null;
     final box = findRenderObject()! as RenderBox;
     final Offset tapPos = box.globalToLocal(globalPosition);
-    final double relative = tapPos.dx / box.size.width;
+    final double relative = (tapPos.dx / box.size.width).clamp(0, 1);
     final Duration position = videoDuration * relative;
     return position;
   }
--- a/lib/src/progress_bar.dart
+++ b/lib/src/progress_bar.dart
@@ -58,11 +58,10 @@ class _VideoProgressBarState extends State<VideoProgressBar> {
   }
 
   void _seekToRelativePosition(Offset globalPosition) {
-    final relativePosition = context.calcRelativePosition(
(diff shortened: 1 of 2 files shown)


koutaro-masaki:
That is not related to the changes in this pull request. It appears that there is a lint issue in the code of the master branch.

diegotori:
Still though. Since I have release authority on this repo, I need to be able to have everything green prior to delaying it to pub.dev.

Once your changes are merged (along with the lint issue), then every other PR will have to re-sync to that lint fix.

lib/src/progress_bar.dart
```diff
@@ -257,14 +259,14 @@ class _ProgressBarPainter extends CustomPainter {
 }
 
 extension RelativePositionExtensions on BuildContext {
-  Duration calcRelativePosition(
+  Duration? calcRelativePosition(
     Duration videoDuration,
     Offset? globalPosition,
   ) {
-    if (globalPosition == null) return Duration.zero;
+    if (globalPosition == null) return null;
```
diegotori, line 266: If you're already returning `Duration.zero` as a default value when calling this method, wouldn't it make sense to keep this line instead of returning `null` here?

diegotori (changes requested):

lib/src/progress_bar.dart
koutaro-masaki: This function is also used [here](https://github.com/fluttercommunity/chewie/blob/d1a99e1967544eca691c67de83038a4e92ad85dd/lib/src/progress_bar.dart#L156), and this property must distinguish between Duration.zero (pointer dragged to the left edge) and null (not dragged).
https://github.com/fluttercommunity/chewie/blob/d1a99e1967544eca691c67de83038a4e92ad85dd/lib/src/progress_bar.dart#L210

diegotori (approved):
LGTM.

diegotori (approved):

PR #810, opened 2024-01-18
[Web] fix exit full screen issue

ToddZeil:
https://github.com/fluttercommunity/chewie/issues/790
https://github.com/fluttercommunity/chewie/issues/688

When viewing full screen on web, returning from full screen causes original video to lose picture.
We re initialise controllers for web only when returning from full screen 

Also refactor un needed set states, refactor play if finished( as video player restarts if finished)

### DIFF
--- a/lib/src/material/material_desktop_controls.dart
+++ b/lib/src/material/material_desktop_controls.dart
@@ -485,40 +485,37 @@ class _MaterialDesktopControlsState extends State<MaterialDesktopControls>
   void _onExpandCollapse() {
     setState(() {
       notifier.hideStuff = true;
+    });
 
-      chewieController.toggleFullScreen();
-      _showAfterExpandCollapseTimer =
-          Timer(const Duration(milliseconds: 300), () {
-        setState(() {
-          _cancelAndRestartTimer();
-        });
+    chewieController.toggleFullScreen();
+
+    _showAfterExpandCollapseTimer =
+        Timer(const Duration(milliseconds: 300), () {
+      setState(() {
+        _cancelAndRestartTimer();
       });
     });
   }
 
   void _playPause() {
-    final isFinished = _latestValue.position >= _latestValue.duration;
-
-    setState(() {
-      if (controller.value.isPlaying) {
+    if (controller.value.isPlaying) {
+      setState(() {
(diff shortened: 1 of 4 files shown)


diegotori:
@ToddZeil thanks for your pending changes.

Definitely like the `setState` refactoring, since that will definitely go towards fixing this long term.

> refactor play if finished( as video player restarts if finished)

This should actually be tied to `ChewieController.looping`, which calls out to `VideoPlayerController.setLooping`. In other words, it should only set the `Duration` back to zero if the underlying `VideoPlayerController` is currently looping.

Once that's done, then that should be good enough for me to approve and merge.

lib/src/material/material_desktop_controls.dart
```diff
@@ -485,40 +485,37 @@ class _MaterialDesktopControlsState extends State<MaterialDesktopControls>
...
-      } else {
-        _cancelAndRestartTimer();
+      });
 
-        if (!controller.value.isInitialized) {
-          controller.initialize().then((_) {
-            controller.play();
-          });
-        } else {
-          if (isFinished) {
```
diegotori, line 515: The `Duration` should be set back to `zero` if the video is finished player and `controller.value.isLooping` is `true`.

diegotori (changes requested):

lib/src/material/material_desktop_controls.dart
ToddZeil: Even with the `VideoPlayerController` `play()` seeking to `zero`? (this may be an update)

image

diegotori: I think I see your point. When hitting play, it will automatically start from the beginning if already finished according to the `VideoController`'s implementation.

In that case, add comments instead, describing this behavior so that other devs are aware that `VideoController` behaves this way.

diegotori (changes requested):

ToddZeil:
@diegotori fair call, added those comments now

ToddZeil:
@diegotori are you happy with the added comments ?

diegotori (approved):
LGTM.

Note that I'll merge and publish this tomorrow, since I don't have my development computer in front of me.

PR #815, opened 2024-02-14
Fix the Safe area conflict for material controls in Android

MadGeorge:
### The issue:

When a player widget is used inside a scaffold with `extendBodyBehindAppBar` set to **true** on a _Scaffold_, the progress bar and the fullscreen button go out of the player bounds which breaks the layout and makes them inaccessible for a user's input.

## Example from real app:
`extendBodyBehindAppBar: true`, controls are falling down

`extendBodyBehindAppBar: false`, controls are fine, but we need the content of the screen go under the app bar

`extendBodyBehindAppBar: true`, and the fix from PR
This is how it should work. 

If I got it right the issue is Android/Fucsia related, since MaterialControls do not used for others. This was not tested on Fucsia.

### DIFF
--- a/lib/src/material/material_controls.dart
+++ b/lib/src/material/material_controls.dart
@@ -261,6 +261,7 @@ class _MaterialControlsState extends State<MaterialControls>
           bottom: !chewieController.isFullScreen ? 10.0 : 0,
         ),
         child: SafeArea(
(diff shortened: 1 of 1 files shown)


diegotori (approved):
LGTM.

PR #823, opened 2024-03-30
Restore Chosen Playback Speed After Seeking

ganeshh123:
When seeking using the progress bar, the playback speed is reset to 1.

I've changed the code to restore the playback speed after seeking to the new position in the progress bar.


diegotori (changes requested):

ganeshh123:
Thanks, I tried that code, but for some reason the speed resets again. Maybe the 1 second wait is necessary?

PR #826, opened 2024-04-03
Add barHeight and handleHeight for MaterialVideoProgressBar

yujune:
### Changes being made

1. Export MaterialVideoProgressBar widget for ChewieController's customControls usage to achieve below ui.
image

2. Add barHeight and handleHeight properties in MaterialVideoProgressBar widget for height customization.

### DIFF
--- a/lib/src/progress_bar.dart
+++ b/lib/src/progress_bar.dart
@@ -9,6 +9,7 @@ class VideoProgressBar extends StatefulWidget {
     this.onDragEnd,
     this.onDragStart,
     this.onDragUpdate,
+    this.draggableProgressBar = true,
     super.key,
     required this.barHeight,
     required this.handleHeight,
@@ -24,6 +25,7 @@ class VideoProgressBar extends StatefulWidget {
   final double barHeight;
   final double handleHeight;
   final bool drawShadow;
+  final bool draggableProgressBar;
 
   @override
   // ignore: library_private_types_in_public_api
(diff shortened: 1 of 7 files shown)


lib/src/progress_bar.dart
```diff
@@ -77,7 +78,7 @@ class _VideoProgressBarState extends State<VideoProgressBar> {
       ),
     );
 
-    return chewieController.draggableProgressBar
+    return widget.draggableProgressBar
```
Ahmadre, line 81: Is there a reason you introduce a new parameter instead of passing the user defined value into the chewieController so that the value is accessable through all of the plugin?

diegotori: Seconded. Everything should revisit l revolve around the `ChewieController`.

diegotori (changes requested):
Please address what @Ahmadre called out. Thanks.

lib/src/progress_bar.dart
yujune: The intension is to use `MaterialVideoProgressBar` widget without depending on the `ChewieControllerProvider`. Hence, other video package like video_player, video_editor can be used for displaying the video and`MaterialVideoProgressBar` for custom controls.

Below is the sample code of using `MaterialVideoProgressBar` with other package.
```dart
//with video_editor
Stack(
      fit: StackFit.expand,
      children: [
        CropGridViewer.preview(
          controller: widget.controller,
        ),
        MaterialVideoProgressBar(
          widget.controller.video,
        ),
      ],
    ),

//with video_player
 Stack(
      fit: StackFit.expand,
      children: [
        VideoPlayer(widget.videoController),
        MaterialVideoProgressBar(
          widget.videoController,
        ),
      ],
    ),
```
I do notice `draggableProgressBar` property is added but no value is being passed, hence, I have added some changes in .

diegotori (approved):
LGTM.

Ahmadre (approved):

PR #828, opened 2024-04-12
Fix the logic of the Center Play Button icon selection

EmreDET:
Hello, this is a small fix. We saw that the plugin has a logical error whenever we did not initialize the video, the icon for the play button would show the "reload"-icon. This happens because: `_latestValue.position >= _latestValue.duration` indicates, that if duration was 0, the initial position would also be 0, therefore the video player icon logic falsely assumed that the video needs to be "reloaded" from the beginning. With this simple fix `&& _latestValue.duration.inSeconds > 0` we ensure, that it should only be treated as "isFinished" if the duration is non-zero, therefore was initialized. If needed, we can even do `&& _latestValue.duration.inMilliseconds > 0` or `_latestValue.duration.inMicroseconds > 0`, for further precision. 
Please correct me if I have made a mistake.

Thank you for your time and efforts.
Emre Y.


MrJohnDev:
v

Good!

EmreDET:
Will this be merged soon? :)

In the meantime for others, here is my fork: https://github.com/EmreDET/chewie

```yaml
chewie:
    git:
      url: https://github.com/EmreDET/chewie.git
      ref: master
```

diegotori:
@EmreDET Thank you for your contribution. 

Please confirm that this issue also applies to Cupertino as well. If it does, then apply the same fix there as well. 

The earliest I'll be able to deploy this will be on Monday.

EmreDET:
Hey @diegotori, thank you for your quick response.
Indeed, the same behavior was falsely in `CupertinoControls` & `MaterialDesktopControls`, I fixed them as well.
I also recreated the iOS example app, so it runs on new devices (iPhone 15 Pro, iOS 17.4).
Lastly, I added a test for that exact behavior.

Please let me know if I did something wrong. Thank you.
Emre Y.

diegotori (approved):
LGTM.

diegotori:
@EmreDET published in version `1.8.3`.

PR #830, opened 2024-04-17
Hide options without play stop button

enesthedev:
This feature based on specific usage of MaterialControls from Chewie, appends hideControlsWithoutPlayButton option to MaterialControls for hide controls without play/pause button of player.


maherjaafar (commented):
LGTM!

PR #841, opened 2024-07-03
Packages upgrades

vaishnavi-2301:
Update video_player dependency to version 2.9.1


diegotori:
@vaishnavi-2301 Thanks for your contribution.

Looks like there are a lot of changes beyond just upgrading `video_player`.

Please isolate the changes to just the ones necessary for upgrading that dependency. Thanks.

Issue #618, opened 2022-03-08
pop wrong page


aslamibnabdulla:
when pressed on speed controls in bottomsheet it pops the main page and keeps the bottomsheet  (not fullscreen  mode).same problem additional options  .  video source cannot be changed   in fullscreen mode.

graphicbeacon:
Experiencing this same issue too

diegotori:
@aslamibnabdulla or @graphicbeacon can you please create a reproducible example project so that I can further diagnose this issue? Thanks in advance.

J-Dark:
Same issue here.
I'm using [Routemaster](https://pub.dev/packages/routemaster) to handle navigation.
Not able to replicate without Routemaster package or Navigator 2.0.
To replicate just copy the simple example from routemaster and push a page with Chewie in it.
I can create a GitHub project if needed.
Also if `fullScreenByDefault: true`, you need to press the exit fullscreen button twice to actually exit fullscreen.

diegotori:
@J-Dark thank you for creating a reproducible example. Hopefully sometime this week, I'll have to take a look.

J-Dark:
@diegotori, any news on this and #647?
Also, this doesn't happen on iOS; only Android and Web.

diegotori:
@J-Dark a reproducible example will help me out tremendously, since it'll allow me to point to a local copy of Chewie and step through the code.

diegotori:
I'll end up taking a look, possibly this week. @maherjaafar if you also want to take a look, go for it as well.

sanchez-gonzalo94:
Hi! Is there an update on this ? I'm experiencing the same issue on version 1.3.4.

hamidwaezi:
Hi! Is there an update on this ? the same issue on version 1.3.6. it pop when open speed control.

J-Dark:
@hamidwaezi I ended up creating `custom_controls.dart` by coping `material_controls.dart` and removing the `Navigator.pop(context);` in `_buildOptionsButton()`.
It's not quite the right behaviour but it's better than popping the main page.

hamidwaezi:
> @hamidwaezi I ended up creating `custom_controls.dart` by coping `material_controls.dart` and removing the `Navigator.pop(context);` in `_buildOptionsButton()`. It's not quite the right behaviour but it's better than popping the main page.

 i just remove this line in _buildOptionsButton
 `useRootNavigator: chewieController.useRootNavigator,`
 and it work like a charm

PR #867, opened 2024-10-28
Fixed pop the wrong page when changing the speed and added support for multiple resolutions

akmalova:
**1. Fixed pop the wrong page when changing the speed**
Issue: https://github.com/fluttercommunity/chewie/issues/618

**2. Added support for multiple resolutions**
Issue: https://github.com/fluttercommunity/chewie/issues/852

Now, to be able to change the video resolution, you need to pass the following parameters to the `ChewieController`:
1. `allowQualityChanging: true` 
2. Map ```
qualities: {
        '240p': '<your link to a 240p video>',
        '360p': '<your link to a 360p video>',
        '720p': <your link to a 720p video>',
        '1080p': '<your link to a 1080p video>',
      }
```. The _key_ can be any string that you want to display in the list of resolutions in the player. The _value_ should be link to the video of this resolution.

Also, in order to be able to change the video resolution, I need to redefine the `videoPlayerController` inside the `ChiewieController`. Therefore, the `onVideoControllerChange` callback has been added, which will allow you to get a new controller when it is replaced.

The quality change has been added for iOS and Android only.

### DIFF
--- a/example/lib/app/app.dart
+++ b/example/lib/app/app.dart
@@ -41,9 +41,10 @@ class _ChewieDemoState extends State<ChewieDemo> {
   }
 
   List<String> srcs = [
-    "https://assets.mixkit.co/videos/preview/mixkit-spinning-around-the-earth-29351-large.mp4",
-    "https://assets.mixkit.co/videos/preview/mixkit-daytime-city-traffic-aerial-view-56-large.mp4",
-    "https://assets.mixkit.co/videos/preview/mixkit-a-girl-blowing-a-bubble-gum-at-an-amusement-park-1226-large.mp4"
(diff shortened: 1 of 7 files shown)


diegotori:
@akmalova Thanks for your contribution.

I do have a major concern with changing the resolution. Since this is driven primarily by `VideoPlayerController`, shouldn't it be relegated to re-creating the `ChewieContoller` instead of internally mutating the `VideoPlayerController`? Just trying to wrap my head regarding this change, because the change that you're proposing represents a fundamental shift in the way `ChewieController` behaves.

@brianegan and/or @Ahmadre, I would like your thoughts on the above.

Also, for the fix that addresses #618, if you can please split that one up into a separate PR, that would be easier for me to review and eventually approve, assuming any concerns that I may raise are eventually addressed.

Thanks in advance.

PR #869, opened 2024-11-06
Update README.md video_player import

AlexanderThiele:
add video_player import otherwise the VideoPlayerController is not available.


diegotori (approved):
LGTM.

PR #871, opened 2024-11-19
Fixed pop the wrong page when changing the speed

akmalova:
**Fixed pop the wrong page when changing the speed**
Issue: https://github.com/fluttercommunity/chewie/issues/618

This was during the speed change and the use of additional options. This was due to the use of the wrong context. In order for the bottom sheet to close correctly, we need to use the context of the bottom sheet itself.
```
await showModalBottomSheet<OptionItem>(
              context: context,
              isScrollControlled: true,
              useRootNavigator: chewieController.useRootNavigator,
              builder: (context) => OptionsDialog(
                options: options,
                options: _buildOptions(context),
                cancelButtonText:
                    chewieController.optionsTranslation?.cancelButtonText,
              ),
```

In this regard, I had to add the `BuildContext context` parameter to the `OptionItem` model in `onTap` callback. Now there is no such problem.
`Function(BuildContext context)? onTap;`

Such a bug was noticed in a project where we use AutoRoute package, so maybe this bug has something to do with using this package.

### DIFF
--- a/lib/src/cupertino/widgets/cupertino_options_dialog.dart
+++ b/lib/src/cupertino/widgets/cupertino_options_dialog.dart
@@ -24,7 +24,7 @@ class _CupertinoOptionsDialogState extends State<CupertinoOptionsDialog> {
         actions: widget.options
             .map(
               (option) => CupertinoActionSheetAction(
-                onPressed: () => option.onTap!(),
+                onPressed: () => option.onTap!(context),
                 child: Text(option.title),
               ),
             )
--- a/lib/src/cupertino/widgets/cupertino_options_dialog.dart
+++ b/lib/src/cupertino/widgets/cupertino_options_dialog.dart
@@ -24,7 +24,7 @@ class _CupertinoOptionsDialogState extends State<CupertinoOptionsDialog> {
         actions: widget.options
             .map(
               (option) => CupertinoActionSheetAction(
-                onPressed: () => option.onTap!(context),
+                onPressed: () => option.onTap(context),
                 child: Text(option.title),
               ),
             )
--- a/lib/src/material/widgets/options_dialog.dart
+++ b/lib/src/material/widgets/options_dialog.dart
@@ -28,7 +28,7 @@ class _OptionsDialogState extends State<OptionsDialog> {
             itemCount: widget.options.length,
             itemBuilder: (context, i) {
               return ListTile(
-                onTap: widget.options[i].onTap,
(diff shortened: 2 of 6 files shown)


diegotori (changes requested):
Other than this nit, everything else looks good for this PR.

akmalova:
Done

lib/src/cupertino/widgets/cupertino_options_dialog.dart
```diff
@@ -24,7 +24,7 @@ class _CupertinoOptionsDialogState extends State<CupertinoOptionsDialog> {
         actions: widget.options
             .map(
               (option) => CupertinoActionSheetAction(
-                onPressed: () => option.onTap!(),
+                onPressed: () => option.onTap!(context),
```
diegotori, line 27: Please modify `OptionItem.onTap` so that it's not null:

```dart
onPressed: () => option.onTap(context),
```

lib/src/models/option_item.dart
```diff
@@ -8,13 +8,13 @@ class OptionItem {
     this.subtitle,
   });
 
-  Function()? onTap;
+  Function(BuildContext context)? onTap;
```
diegotori, line 11: Two things:

1. Make all of this class's properties `final`.
2. Rename this function to:

```dart
final void Function(BuildContext context) onTap;
```

lib/src/material/widgets/options_dialog.dart
```diff
@@ -28,7 +28,7 @@ class _OptionsDialogState extends State<OptionsDialog> {
             itemCount: widget.options.length,
             itemBuilder: (context, i) {
               return ListTile(
-                onTap: widget.options[i].onTap,
+                onTap: () => widget.options[i].onTap?.call(context),
```
diegotori, line 31: If you make `OptionItem.onTap` non-null, then you won't need to do this.

diegotori (changes requested):

diegotori:
@akmalova Please run `dart format` on the current changes. Also added some additional feedback. Please address that in order for me to approve your changes. Thanks.

diegotori:
Lastly, please re-sync your changes with `master`. Thanks.

diegotori:
@akmalova you still with us on this PR?

akmalova:
@diegotori Yes, I'm sorry, I didn't have time. I've fixed everything now.

diegotori (approved):
LGTM.

PR #875, opened 2024-12-15
Add background tap to pause video feature

Ortes:
When the video play and the feature enabled one click anywhere on the overlay pause the video

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -308,6 +308,7 @@ class ChewieController extends ChangeNotifier {
     this.progressIndicatorDelay,
     this.hideControlsTimer = defaultHideControlsTimer,
     this.controlsSafeAreaMinimum = EdgeInsets.zero,
+    this.pauseOnBackgroundTap = false,
   }) : assert(
           playbackSpeeds.every((speed) => speed > 0),
           'The playbackSpeeds values must all be greater than 0',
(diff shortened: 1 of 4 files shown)


maherjaafar (commented):
Can you add a video to demonstrate how this feature works?

Ortes:
Sure when clicking anywhere on the video the video play/pause instead of just making the overlay appear

https://github.com/user-attachments/assets/6e4acf92-97e2-46d8-8b6f-0f745a3ebe4d

Ortes:
> @Ortes please fix the formatting issues as per the CI results. Thanks.

Is this ok ?

diegotori (approved):
LGTM.

diegotori:
> It might be a good idea to rename the parameter if `pauseOnBackgroundTap` is unclear

Up to you.

Ortes:
I would rather have this as the default behavior as it is the behavior on Youtube and a lot of other players out here. But I understand you don't want breaking changes

diegotori:
> I would rather have this as the default behavior as it is the behavior on Youtube and a lot of other players out here. But I understand if you don't want breaking changes

Yeah, it should be well documented for now that this option is available. In a future minor update, we'll turn it on by default.

I guess your task right now is to document the hell out of this for the time being. Thanks.

diegotori (approved):
LGTM.

PR #884, opened 2025-01-09
fix: remove default case in AdaptiveControls and correct opacity valu…

Stanely254:
…e in MaterialControls


diegotori:
@Stanely254 I believe this was already fixed. Please re-sync your codebase with the current master version. Thanks.

PR #896, opened 2025-02-27
Fixed allowMute being ignored on Desktop

mpoimer:


mpoimer:
@diegotori done

lib/src/material/material_desktop_controls.dart
```diff
@@ -31,8 +31,7 @@ class MaterialDesktopControls extends StatefulWidget {
   }
 }
 
-class _MaterialDesktopControlsState extends State<MaterialDesktopControls>
-    with SingleTickerProviderStateMixin {
+class _MaterialDesktopControlsState extends State<MaterialDesktopControls> with SingleTickerProviderStateMixin {
```
diegotori, line 34: You may want to revert the formatting changes, so that the CI doesn't complain when it attempts to format it.

That and it'll make it easier to see the actual change made.

diegotori (changes requested):

lib/src/material/material_desktop_controls.dart
mpoimer: done

diegotori: Looks like there's still some formatting issues that CI is complaining about.

To fix those, downgrade to 3.27, then run dart formal in the codebase. Thanks in advance.

mpoimer: done

Issue #909, opened 2025-04-01
The sliding progress bar always displays the loading component


wangya123456789:
I know this problem has nothing to do with the chewie library, is there any way to solve this problem?

C:\Users\Admin>flutter --version
Flutter 3.27.1 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 17025dd882 (4 months ago) • 2024-12-17 03:23:09 +0900
Engine • revision cb4b5fff73
Tools • Dart 3.6.0 • DevTools 2.40.2

https://github.com/user-attachments/assets/58f3d099-1729-4b32-b209-1770313cde64

rajatkumar-patel:
I am facing same issue

wangya123456789:
> [王雅123456789](https://github.com/wangya123456789)我遇到了同样的问题，您能找到任何解决方法吗？

 _chewieController = ChewieController(
          videoPlayerController: _videoPlayerController,
          progressIndicatorDelay:
              PlatformProvider.isAndroid() ? const Duration(days: 1) : null,
        );

This method can solve

Issue #906, opened 2025-03-30
Video not playable after completion due to infinite loading when loop is disabled


naghinezhad:
When playing a video using the Chewie package for the first time, the video plays until the end without any issues. However, after the video completes, an infinite loading indicator appears, and the video cannot be played again. This issue occurs **when the loop feature is disabled**.  

**Steps to reproduce:**  
1. Play a video using Chewie.  
2. Let the video play until completion.  
3. Try to play the video again after it finishes.  

**Expected behavior:**  
The video should be playable again after completion, even when the loop is disabled.  

**Actual behavior:**  
An infinite loading indicator appears after the video completes, and the video cannot be played again.  

**Environment:**  
- Flutter version: 3.29.2  
- Chewie version: 1.11.0  
- Platform: Android only

HeropolisDa2ny:
Yep I am encountering the same problem and the infinite loops seems appearing on others features : https://github.com/fluttercommunity/chewie/issues/907

naghinezhad:
What is the solution?

yann-elpyoo:
@naghinezhad you can try this workaround 
https://github.com/fluttercommunity/chewie/issues/907#issuecomment-2773601308

naghinezhad:
@timoxd7 Is there a problem with the pull request? #912

Issue #903, opened 2025-03-22
Play Pause and 10sec forward backword option not appearing once seek to a position or clicking on 10sec forward backword buttons.


saiful00000:
Once its start buffering when seeking the progress of by clicking 10sec forward or backward button its keep showing the Circular Progress Indicator even after video play completed.

Reproducible code:

`import 'package:chewie/chewie.dart';
import 'package:flutter/material.dart';
import 'package:video_player/video_player.dart';

class MovieVideoPlayer extends StatefulWidget {
  final String? url;

  const MovieVideoPlayer({super.key, required this.url});

  @override
  State<MovieVideoPlayer> createState() => _MovieVideoPlayerState();
}

class _MovieVideoPlayerState extends State<MovieVideoPlayer> {
  late final VideoPlayerController playerController;
  late final ChewieController chewieController;

  @override
  void initState() {
    playerController = VideoPlayerController.networkUrl(
      Uri.parse(widget.url ?? 'http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4'),
    );
    chewieController = ChewieController(
      videoPlayerController: playerController,
      aspectRatio: 16 / 9,
      allowedScreenSleep: false,
      allowFullScreen: true,
    );

    super.initState();
  }

  @override
  void dispose() {
    playerController.dispose();
    chewieController.dispose();

    super.dispose();
  }

  @override
  Widget build(BuildContext context) {
    if (widget.url == null) {
      return const Center(
        child: Text('No video found.'),
      );
    }
    return AspectRatio(
      aspectRatio: 16 / 9,
      child: Chewie(
        controller: chewieController,
      ),
    );
  }
}
`

dharmic6:
to add more context, this happens in case of listview with videos.

```
  video_player: 2.9.3
  chewie: 1.11.0
```

PR #912, opened 2025-04-02
Add workaround for invalid buffering info on android

timoxd7:
## Info
As already described in multiple issues, there is currently a problem with android reporting "isBuffering" with true even if the video is already playing, preventing the video controls to be displayed. This is a workaround, checking (on android only) if the video is actually buffering. Tested with a network- and file-based video.

>NOTE: This is only a workaround until video_player and flutter is fixed. See https://github.com/flutter/flutter/issues/165149 for more information. Also thanks to @wenwangwill who added a workaround as comment i got inspired from.

## Additional Issues not solved
If pausing the video and seeking to some point (using seekTo, so clicking on the progress bar) still leads to the infinit loading bar, as no callback is called by the video controller and thus the buffering not updated.

As workaround, just start playing on seekTo like this:

```dart
  // Your init code can be above
  videoController.addListener(yourListeningMethod);

  // ...

  bool wasPlayingBefore = false;
  void yourListeningMethod() {
    if (!videoController.value.isPlaying && !wasPlayingBefore) {
      // -> Workaround if seekTo another position while it was paused before
      //    on Android this might lead to infinite loading, so just play the
      //    video again.
      videoController.play();
    }

    wasPlayingBefore = videoController.value.isPlaying;

  // ...
  }

```

## Related to
#909
#907 
#906
#903

### DIFF
--- a/lib/src/cupertino/cupertino_controls.dart
+++ b/lib/src/cupertino/cupertino_controls.dart
@@ -1,5 +1,4 @@
 import 'dart:async';
-import 'dart:io';
 import 'dart:math' as math;
 import 'dart:ui' as ui;
 
@@ -811,33 +810,7 @@ class _CupertinoControlsState extends State<CupertinoControls>
   void _updateState() {
     if (!mounted) return;
 
-    late final bool buffering;
-
-    if (Platform.isAndroid) {
-      if (controller.value.isBuffering) {
-        // -> Check if we actually buffer, as android has a bug preventing to
-        //    get the correct buffering state from this single bool.
-        final VideoPlayerValue value = controller.value;
-
-        final int position = value.position.inMilliseconds;
-
-        // Special case, if the video is finished, we don't want to show the
-        // buffering indicator anymore
-        if (position >= value.duration.inMilliseconds) {
-          buffering = false;
-        } else {
-          final int buffer =
-              value.buffered.lastOrNull?.end.inMilliseconds ?? -1;
-
-          buffering = position >= buffer;
-        }
-      } else {
-        // -> No buffering
-        buffering = false;
-      }
-    } else {
-      buffering = controller.value.isBuffering;
-    }
+    final bool buffering = getIsBuffering(controller);
 
     // display the progress bar indicator only after the buffering delay if it has been set
     if (chewieController.progressIndicatorDelay != null) {
--- a/lib/src/cupertino/cupertino_controls.dart
+++ b/lib/src/cupertino/cupertino_controls.dart
@@ -818,10 +818,19 @@ class _CupertinoControlsState extends State<CupertinoControls>
         // -> Check if we actually buffer, as android has a bug preventing to
         //    get the correct buffering state from this single bool.
         final VideoPlayerValue value = controller.value;
-        final int buffer = value.buffered.lastOrNull?.end.inMilliseconds ?? -1;
+
         final int position = value.position.inMilliseconds;
 
-        buffering = position >= buffer;
+        // Special case, if the video is finished, we don't want to show the
+        // buffering indicator anymore
+        if (position >= value.duration.inMilliseconds) {
+          buffering = false;
+        } else {
+          final int buffer =
+              value.buffered.lastOrNull?.end.inMilliseconds ?? -1;
+
+          buffering = position >= buffer;
+        }
       } else {
         // -> No buffering
         buffering = false;
--- a/lib/src/cupertino/cupertino_controls.dart
+++ b/lib/src/cupertino/cupertino_controls.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
 import 'dart:math' as math;
 import 'dart:ui' as ui;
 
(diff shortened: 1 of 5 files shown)


lib/src/cupertino/cupertino_controls.dart
```diff
@@ -810,9 +811,37 @@ class _CupertinoControlsState extends State<CupertinoControls>
...
+
+          buffering = position >= buffer;
+        }
+      } else {
+        // -> No buffering
+        buffering = false;
+      }
+    } else {
+      buffering = controller.value.isBuffering;
+    }
```
diegotori, line 840: Encapsulate this into a private function.

lib/src/material/material_controls.dart
```diff
@@ -645,9 +646,37 @@ class _MaterialControlsState extends State<MaterialControls>
...
+
+          buffering = position >= buffer;
+        }
+      } else {
+        // -> No buffering
+        buffering = false;
+      }
+    } else {
+      buffering = controller.value.isBuffering;
+    }
```
diegotori, line 675: Encapsulate this into a private function.

lib/src/material/material_desktop_controls.dart
```diff
@@ -581,9 +582,37 @@ class _MaterialDesktopControlsState extends State<MaterialDesktopControls>
...
+
+          buffering = position >= buffer;
+        }
+      } else {
+        // -> No buffering
+        buffering = false;
+      }
+    } else {
+      buffering = controller.value.isBuffering;
+    }
```
diegotori, line 611: Encapsulate this into a private function.

diegotori (changes requested):

yann-elpyoo:
For information until the bug is fixed, **and for chewie users only** : it is possible to use a simple and complete workaround.

The thing is to add a long duration for the loader to appear. Then nothing is blocked anymore even at the end of the video! No need for looping too. It just shows the video paused and played each time seek to is called... 
```
          progressIndicatorDelay:
              PlatformProvider.isAndroid() ? const Duration(days: 1) : null,
        );
```

For non chewie users suffering from it, maybe wrapping with chewie could be a temporary solution...

naghinezhad:
@timoxd7 @diegotori
Please review and merge this pull request so the related bug can be resolved. We are waiting for it.

diegotori (changes requested):

timoxd7:
@naghinezhad i am a bit in two minds about this pr. I will incorporate the requested changes, but it still is not a complete fix. 

I don't know if it works good with live streams. As @yann-elpyoo wrote, you _could_ fix it by disabling indirectly the loading spinner for android, but this also removes the feedback to the user that something is buffering.

@diegotori i updated according to your review. I also tested @yann-elpyoo solution and added it to the readme, both work together too (as my workaround will just be _hidden_ by disabling the loader anyway).

As maintainer you can decide wether to merge this, as the root problem is not relying in chewie itself. However, we now have it documented here at least and if someone needs a fix, can use it util it is fixed in flutter itself.

lib/src/helpers/utils.dart
```diff
@@ -1,3 +1,7 @@
+import 'dart:io';
```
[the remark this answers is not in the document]
diegotori: Instead, have that method take in a `BuildContext` and then use `(Theme.of(context).platform` to find the current platform.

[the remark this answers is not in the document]
timoxd7: Will fix it!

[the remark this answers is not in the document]
timoxd7: Just saw your second comment. Should i change it still to `BuildContext`?

diegotori (changes requested):
For the new utility method, add comments related to the issue your changes are trying to address. 

Thanks.

lib/src/helpers/utils.dart
```diff
@@ -34,6 +34,11 @@ String formatDuration(Duration position) {
   return formattedTime;
 }
 
+/// Get the current buffering state of the video player. Will invoke a
+/// Workaround for Android, as a bug in the video_player plugin prevents to get
+/// the actual buffering state, always returning true and breaking ui elements.
+/// For this, the actual buffer position is used to determine if the video is
+/// buffering or not. See #912 for more details.
```
diegotori, line 41: Change this to:

```dart
/// Gets the current buffering state of the video player. 
/// 
/// For Android, it will use a workaround due to a [bug](https://github.com/flutter/flutter/issues/165149) 
/// affecting the `video_player` plugin, preventing it from getting the 
/// actual buffering state. This currently results in the `VideoPlayerController` always buffering,
/// thus breaking UI elements.
///
/// For this, the actual buffer position is used to determine if the video is
/// buffering or not. See Issue [#912](https://github.com/fluttercommunity/chewie/pull/912) for more details.
```

diegotori (changes requested):

diegotori:
@timoxd7 other than the final nit that I called out, the rest LGTM.

diegotori (approved):
LGTM.

PR #932, opened 2025-09-22
Add feature double tap for fullscreen toggle

Ortes:


Ortes:
Closing in favor of #964, which implements the full YouTube gesture policy: double-tap toggles fullscreen on the desktop controls (where YT has it), while the mobile hit area's double-tap is reserved for ±10s seek zones. Same `allowDoubleTapToggleFullScreen` option, relocated.

PR #936, opened 2025-12-23
Added an option to use the native fullscreen of a browser in webapps

JuggerbAUT:
This PR introduces native web fullscreen support to provide a more immersive video playback experience in web browsers.

### Problem
Currently, the fullscreen mode in chewie on web has some limitations:

- **Browser UI remains visible:** When entering fullscreen, browser elements like the address bar and tabs are still visible, which is not a true fullscreen experience.
- **Constrained by parent containers:** in a width restriced webapp the fullscreen mode of the player is also restricted -> removing this restriction with a variable set in a provider leads to problems with the player notifier

### Solution
 `useNativeWebFullscreen` flag: A new boolean flag, `useNativeWebFullscreen`, has been added to `ChewieController`. This flag is true by default. When set to false, it reverts to the original fullscreen behavior.

### DIFF
--- a/lib/src/helpers/fullscreen_controller.dart
+++ b/lib/src/helpers/fullscreen_controller.dart
@@ -1,1 +1,1 @@
-export 'fullscreen_controller_unsupported.dart' if (dart.library.html) 'fullscreen_controller_web.dart';
+export 'fullscreen_controller_unsupported.dart' if (dart.library.js_interop) 'fullscreen_controller_web.dart';
--- a/lib/src/helpers/fullscreen_controller.dart
+++ b/lib/src/helpers/fullscreen_controller.dart
@@ -1,0 +1,1 @@
+export 'fullscreen_controller_unsupported.dart' if (dart.library.html) 'fullscreen_controller_web.dart';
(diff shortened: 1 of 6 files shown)


lib/src/helpers/fullscreen_controller.dart
```diff
@@ -0,0 +1 @@
+export 'fullscreen_controller_unsupported.dart' if (dart.library.html) 'fullscreen_controller_web.dart';
```
diegotori: Do not use `dart.library.html` anymore. 

Instead use `dart.library.js_interop`.

diegotori (changes requested):

lib/src/helpers/fullscreen_controller.dart
JuggerbAUT: Yeah, I've seen it, but a series of unfortunate coincidences made me defer it. I'll take a 2nd look on it next week. Sorry for keeping it open that long and not fixing it.

a1rwulf:
This can be closed, @Ortes has contributed native fullscreen support already.

PR #946, opened 2026-05-31
feat(web): native browser fullscreen via the Fullscreen API

Ortes:
## Problem

On Flutter Web, entering fullscreen only pushes a route that expands the Flutter view **inside the browser window** — it never enters the browser's real, OS-level fullscreen. Users expect the video to fill the whole screen (like every `<video>` element and every other web player), but instead it just fills the page. This is a long-standing rough edge for web users.

## What this does

Adds an opt-out `ChewieController.useNativeFullScreenOnWeb` flag (**default `true`**) that, on web, also drives the browser's [Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API):

- Entering Chewie fullscreen calls `document.documentElement.requestFullscreen()` (issued before the first `await`, so it's still inside the user-gesture handler the browser requires).
- Exiting Chewie fullscreen calls `document.exitFullscreen()`.
- Pressing **Escape** (which the browser handles itself) fires `fullscreenchange`; Chewie listens and exits its own fullscreen route so the two states stay in sync.

No behavior change on Android / iOS / desktop — the flag has no effect off web.

## Implementation notes

- Native fullscreen unavoidably needs JS interop. It's isolated behind a conditional import so non-web targets never reference `package:web`:
  - `lib/src/web_fullscreen.dart` — real impl (`package:web` + `dart:js_interop`)
  - `lib/src/web_fullscreen_stub.dart` — no-op stubs
- `lib/src/chewie_player.dart` — new field + constructor arg + `copyWith`; listener registration in `initState`/`dispose`; enter/exit calls in `_pushFullScreenWidget`, all gated by the flag.
- `pubspec.yaml` — adds `web: ^1.0.0`.

Happy to flip the default to `false` if you'd prefer it strictly opt-in, or to add a small example/integration test — let me know your preference.


lib/src/chewie_player.dart
```diff
@@ -1,6 +1,8 @@
 import 'dart:async';
 
 import 'package:chewie/src/chewie_progress_colors.dart';
+// ignore: uri_does_not_exist
```
diegotori: Don't ignore lint issues. 

Instead, think of a way to abstract this and then use the abstraction here. 

The abstraction should conditionally import the implementation.

lib/src/chewie_player.dart
```diff
@@ -1,6 +1,8 @@
...
+import 'web_fullscreen_stub.dart' if (dart.library.html) 'web_fullscreen.dart';
```
diegotori: Don't use `dart.library.html`.

Instead, you have to use `dart.library.js_interop`.

diegotori (changes requested):

lib/src/chewie_player.dart
Ortes: Done. Removed the lint ignore — `chewie_player.dart` now imports a plain `web_fullscreen.dart` abstraction that conditionally exports the stub vs. the real implementation (moved to `web_fullscreen_impl.dart`).

lib/src/chewie_player.dart
Ortes: Done. The conditional export now uses `dart.library.js_interop` instead of `dart.library.html`.

diegotori:
There are conflicts to resolve before this can be merged. Thanks.

Ortes:
Hey @diegotori — following up on getting maintainer/collaborator access on chewie. I've got 15 open PRs at this point (chapters, subtitle tracks, audio tracks, keyboard shortcuts, fullscreen improvements, etc.) and would be glad to help review/merge others' PRs too if that'd take some load off you. Let me know what you'd need to see from me to move forward on that.

diegotori (approved):
LGTM.

PR #947, opened 2026-05-31
feat: add disableFullScreenRoute to let the embedder own fullscreen

Ortes:
## Motivation

Chewie's fullscreen is implemented by **pushing a dedicated route** and re-parenting the player into it. On web this reparents the platform-view `<video>` element; on exit the original view goes blank, and the common workaround is to re-initialize the controller — which for an HLS source means a full manifest reload and rebuffer.

Some embedders want to own the fullscreen presentation themselves (e.g. expand their own layout and drive the browser's native Fullscreen API) without Chewie's route getting in the way.

## What this adds

`ChewieController.disableFullScreenRoute` (default `false`). When `true`:

- Toggling fullscreen **no longer pushes/pops** Chewie's internal route.
- `isFullScreen` still flips, so the fullscreen control icon updates as usual.
- The player widget stays mounted in place; the host app is responsible for the fullscreen presentation.

This sidesteps the platform-view teardown/reload entirely by never moving the element.

## API

- `bool disableFullScreenRoute` on `ChewieController` (constructor + `copyWith`), default `false`.

## Backwards compatibility

Fully backwards compatible — defaults to `false`, preserving the current route-based behavior. Opt-in only.

## Testing

- `dart analyze` — no issues.
- In production use in an HLS web player, where it eliminates the manifest reload/rebuffer on fullscreen exit.

Complementary to the native web-fullscreen work in #946.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [Unreleased]
+* 🖥️ [#947](https://github.com/fluttercommunity/chewie/pull/947): Add `ChewieController.disableFullScreenRoute` (default `false`). When `true`, toggling fullscreen no longer pushes/pops Chewie's internal fullscreen route — `isFullScreen` still flips so the control icon updates, but the player stays mounted in place and the host app owns the fullscreen presentation. Thanks [Ortes](https://github.com/Ortes).
+
 ## [1.15.0]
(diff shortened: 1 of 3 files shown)


diegotori (approved):
LGTM.

PR #948, opened 2026-05-31
feat: multi-track subtitle selection

Ortes:
## Motivation

Streaming sources (notably HLS) expose multiple **subtitle renditions** in their manifest, but Chewie currently has no UI to let the user pick one. This adds a source-agnostic subtitle-track picker to the Material controls.

## What this adds

- A track picker in the options menu (plus an **Off** entry) that appears only when `subtitleTracks` is non-empty.
- Support for streaming cues that arrive over time (via `setLiveSubtitle`), alongside the existing static `subtitle` cue list.

The widget is **source-agnostic**: the host populates the track list and reacts to selection; Chewie never assumes HLS or any particular backend.

## API (all opt-in — defaults preserve current behavior)

`ChewieController`:
- `List<SubtitleTrack> subtitleTracks` — selectable tracks (default empty).
- `Object? activeSubtitleTrackId` — currently selected id (`null` = Off).
- `void Function(SubtitleTrack?)? onSubtitleTrackChanged` — selection callback.
- `setSubtitleTracks()` / `selectSubtitleTrack()` — for tracks known only after load.
- `setLiveSubtitle(String?)` + `ValueNotifier<String?> liveSubtitle` — push current cue text for streaming sources.
- `bool get hasSubtitleTracks` — gates the menu entry.

New: `SubtitleTrack` model and `SubtitleTrackDialog` bottom sheet.

## Backwards compatibility

Fully backwards compatible — every field is optional and defaults to the current behavior. No menu entry appears unless the host provides tracks.

## Testing

- `dart analyze` — no issues.
- In production use in an HLS web player (subtitle tracks parsed from the manifest, cues pushed via `setLiveSubtitle`).

Independent of the audio-track PR (#949) — single commit, based on `master`.

PR #949, opened 2026-05-31
feat: multi-track audio selection

Ortes:
## Motivation

Streaming sources (notably HLS) expose multiple **audio renditions** (languages, commentary, codecs) in their manifest, but Chewie has no UI to switch between them. This adds a source-agnostic audio-track picker to the Material controls.

## What this adds

- An audio-track picker in the options menu, shown only when **more than one** track is available (a single track offers nothing to choose).
- Unlike subtitles, audio is never "off": one track is always active.

The widget is **source-agnostic**: the host populates the track list and performs the actual rendition switch; Chewie never assumes HLS or any particular backend.

## API (all opt-in — defaults preserve current behavior)

`ChewieController`:
- `List<AudioTrack> audioTracks` — selectable tracks (default empty).
- `Object? activeAudioTrackId` — currently selected id.
- `void Function(AudioTrack)? onAudioTrackChanged` — selection callback.
- `setAudioTracks()` / `selectAudioTrack()` — for tracks known only after load.
- `bool get hasAudioTracks` — true only when `length > 1`; gates the menu entry.

New: `AudioTrack` model and `AudioTrackDialog` bottom sheet.

## Backwards compatibility

Fully backwards compatible — every field is optional and defaults to the current behavior. No menu entry appears unless the host provides more than one track.

## Testing

- `dart analyze` — no issues.
- In production use in an HLS web player (audio renditions parsed from the manifest, switched on selection).

Independent of the subtitle-track PR (#948) — single commit, based on `master`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [Unreleased]
+* 🔊 [#949](https://github.com/fluttercommunity/chewie/pull/949): Add source-agnostic multi-track audio selection to the Material controls, via the new `AudioTrack` model and `ChewieController.audioTracks` / `activeAudioTrackId` / `onAudioTrackChanged`. The picker only appears when more than one track is provided. Thanks [Ortes](https://github.com/Ortes).
+
 ## [1.15.0]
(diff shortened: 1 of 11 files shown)


diegotori:

diegotori:
@Ortes please resolve conflicts. Thanks.

diegotori (approved):
LGTM.

PR #953, opened 2026-07-09
feat: YouTube-style seek indicator for keyboard seeking

Ortes:
## What

Adds a YouTube-style seek indicator that flashes on screen when the user seeks with the keyboard arrows on the desktop controls (`MaterialDesktopControls`).

- Pressing → / ← shows a small pill with a directional icon and the amount seeked.
- Repeated presses in the **same** direction keep the pill visible and **accumulate** the total (e.g. `10s` → `20s` → `30s`), just like YouTube.
- Pressing the opposite direction (or after the pill fades ~0.9s) resets the counter.

## How

- New reusable widget `lib/src/seek_indicator.dart` (fades via `AnimatedOpacity`, mirrors the existing `CenterSeekButton` style; purely visual — wrapped in `IgnorePointer`).
- `_MaterialDesktopControlsState` accumulates the seeked seconds in `_seekForward`/`_seekBackward` (kept in sync with the existing `Duration` seek step) and auto-hides via a `Timer`, mirroring the existing buffering-indicator pattern.
- Gated by a new `ChewieController.showSeekIndicator` option (default `true`), wired through `copyWith`.

## Tests

`test/seek_indicator_test.dart` covers accumulation on repeated presses, direction reset, auto fade-out, the `showSeekIndicator: false` opt-out, and the option default/`copyWith`.

`dart format`, `flutter analyze lib test`, and the full `flutter test` suite are clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [Unreleased]
+* ⏩ [#953](https://github.com/fluttercommunity/chewie/pull/953): Add a YouTube-style seek indicator that flashes the seeked amount when seeking with the keyboard arrows on desktop, accumulating on repeated presses in the same direction (e.g. `10s` → `20s` → `30s`). Configurable via `ChewieController.showSeekIndicator` (default `true`), with the per-press amount set by `ChewieController.keyboardSeekDuration` (default 10 seconds). Thanks [Ortes](https://github.com/Ortes).
+
 ## [1.15.0]
 * 🌐 [#946](https://github.com/fluttercommunity/chewie/pull/946): Web: enter the browser's native (OS-level) fullscreen via the Fullscreen API instead of only expanding the Flutter view inside the browser window. Pressing Escape to leave browser fullscreen also exits Chewie's fullscreen. Controlled by the new `ChewieController.useNativeFullScreenOnWeb` flag (defaults to `true`; no effect on non-web platforms). Thanks [Ortes](https://github.com/Ortes).
 * 🖱️ [#950](https://github.com/fluttercommunity/chewie/pull/950): Show click cursor on hover over Material controls and progress bar. Thanks [Ortes](https://github.com/Ortes).
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -365,6 +365,7 @@ class ChewieController extends ChangeNotifier {
     this.controlsSafeAreaMinimum = EdgeInsets.zero,
     this.pauseOnBackgroundTap = false,
     this.showSeekIndicator = true,
+    this.keyboardSeekDuration = const Duration(seconds: 10),
   }) : assert(
          playbackSpeeds.every((speed) => speed > 0),
          'The playbackSpeeds values must all be greater than 0',
@@ -427,6 +428,7 @@ class ChewieController extends ChangeNotifier {
     routePageBuilder,
     bool? pauseOnBackgroundTap,
     bool? showSeekIndicator,
+    Duration? keyboardSeekDuration,
   }) {
     return ChewieController(
       draggableProgressBar: draggableProgressBar ?? this.draggableProgressBar,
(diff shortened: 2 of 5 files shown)


Ortes:
Hi @diegotori 👋

Beyond this PR, I'd like to offer to help with maintenance on chewie if you're open to it.

A bit of context:
- I've had PRs merged here before (#876 desktop keyboard controls, #910 web example fix), and just opened three more, each with tests, a CHANGELOG entry and green CI: #953 (keyboard seek indicator), #954 (auto-hide cursor in fullscreen), #955 (F-key fullscreen toggle).
- I use chewie **in production at my company**, and also **personally on a daily/weekly basis** inside a custom movie player I built on top of it — so I'm exercising it heavily across **web and mobile**.

I know the open-PR queue is long, and I'd be glad to take some of that load off you: triaging and reviewing PRs, keeping the package current with new Flutter releases, and helping land fixes. Would you be open to adding me as a co-maintainer (or triage access to start)? Happy to do whatever works best for you and the Flutter Community org.

Either way, thanks for maintaining this 🙏

diegotori:
@Ortes please fix the conflicts here. Thanks.

Ortes:
Heads-up: #964 is stacked on this branch — it extracts the indicator state into a shared mixin and reuses `SeekIndicator` for YouTube-style double-tap seeking on mobile. This PR stays reviewable standalone; #964 rebases once this lands.

diegotori:
@Ortes please resolve conflicts. Thanks.

ibrahim-iqbal (approved):
Nice feature — the accumulate-in-same-direction, reset-on-opposite behaviour matches YouTube exactly, and the pill layout at `Alignment(±0.6, 0)` gives the same off-centre placement.

**Widget shape**

- `SeekIndicator` wrapped in `IgnorePointer` — correct, the pill would otherwise soak up clicks on the middle of the video.
- `AnimatedOpacity` with a 300 ms fade + a 900 ms hide-timer is the same rhythm as YouTube's own; long enough to read `30 s` after a triple-tap, short enough not to overlap a subsequent tap sequence.
- `Colors.black.withValues(alpha: 0.55)` — good use of the new `withValues` (the deprecated `withOpacity` would also work but the current API is right).

**Controls plumbing**

- `_seekIndicatorTimer` is cancelled in `dispose` alongside the existing `_hideTimer` / `_initTimer` / `_showAfterExpandCollapseTimer` — no leak.
- `_bumpSeekIndicator` gate on `chewieController.showSeekIndicator` first is right: an embed that disables the feature stays entirely off the setState path, so the widget cost is zero when opted out.
- `keyboardSeekDuration` is threaded into both directions (the hardcoded `Duration(seconds: -10)` / `Duration(seconds: 10)` becomes user-configurable) — nice bonus, plays well with `ChewieController.copyWith` for downstream users.

**Small notes, non-blocking:**

- `_seekIndicatorSeconds` is an `int` while `keyboardSeekDuration` is a `Duration`. If someone sets `keyboardSeekDuration: const Duration(milliseconds: 5500)`, `step` becomes `5` (via `.inSeconds`) and the label reads `5 s / 10 s / …`, quietly under-counting. Documenting `keyboardSeekDuration` as "whole seconds" on the field, or clamping the display to `duration.inMilliseconds ~/ 1000` explicitly, would make that surprise less likely.
- The `Text('$seconds s')` label reads correctly in English but ignores locale (Arabic-Indic digits, RTL `س`, etc.). Fine as-is — chewie doesn't currently localise button/hover strings — but worth a comment if you want to leave a hook for future i18n.
- Two `Alignment(±0.6, 0)` magic values — arguably nicer as `_kIndicatorAlignmentForward` / `_Backward` constants at the top of the file so a designer tweak lands in one place. Nit.

Test file (`seek_indicator_test.dart`, 133 lines) covers the accumulate/reset behaviour end-to-end, which is exactly what a future refactor of the timer logic will need to preserve. LGTM.

diegotori (approved):
LGTM.

PR #954, opened 2026-07-09
feat: auto-hide the mouse cursor while idle in fullscreen

Ortes:
## What

In fullscreen the mouse cursor currently stays visible forever. This makes it auto-hide together with the controls once the player goes idle, and reappear as soon as the mouse moves — the behaviour of most video players (YouTube, VLC, …).

## How

- Each controls widget already wraps the whole video surface in a root `MouseRegion` whose `onHover` reveals the controls. We simply set that region's `cursor` to `SystemMouseCursors.none` while `hideCursorInFullScreen && isFullScreen && hideStuff`, and `MouseCursor.defer` otherwise (so inner per-widget cursors still win when the controls are visible).
- Applied to all three control sets (`MaterialDesktopControls`, `MaterialControls`, `CupertinoControls`), matching the cross-file precedent of `pauseOnBackgroundTap`.
- No new timer or platform channel: `SystemMouseCursors.none` maps to CSS `cursor: none` on web and works on desktop too; it is a no-op on pointerless devices.
- Gated by a new `ChewieController.hideCursorInFullScreen` option (default `true`), wired through `copyWith`.

## Tests

`test/hide_cursor_test.dart` covers the option default/`copyWith` and that the cursor is never hidden while **not** in fullscreen even when the controls are idle (the gate).

`dart format`, `flutter analyze lib test`, and the full `flutter test` suite are clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [Unreleased]
+* 🖱️ [#954](https://github.com/fluttercommunity/chewie/pull/954): Auto-hide the mouse cursor along with the controls while idle in fullscreen, and show it again on mouse movement (web and desktop). Configurable via `ChewieController.hideCursorInFullScreen` (default `true`). Thanks [Ortes](https://github.com/Ortes).
+
 ## [1.15.0]
 * 🌐 [#946](https://github.com/fluttercommunity/chewie/pull/946): Web: enter the browser's native (OS-level) fullscreen via the Fullscreen API instead of only expanding the Flutter view inside the browser window. Pressing Escape to leave browser fullscreen also exits Chewie's fullscreen. Controlled by the new `ChewieController.useNativeFullScreenOnWeb` flag (defaults to `true`; no effect on non-web platforms). Thanks [Ortes](https://github.com/Ortes).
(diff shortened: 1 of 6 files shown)


diegotori:

ibrahim-iqbal (approved):
Clean. `_idleCursor` reads the three signals (`hideCursorInFullScreen`, `isFullScreen`, `hideStuff`) and returns `SystemMouseCursors.none` only when all three align, otherwise `MouseCursor.defer` — which is the right value for "let whatever's underneath decide", so the cursor stays whatever the app was using and doesn't get frozen to `basic`.

Reusing the existing root `MouseRegion` (the same one whose `onHover` already resets the controls-hide timer) means the cursor comes back the instant the user moves the mouse — no separate listener needed. Nice.

One small note: the `_idleCursor` getter is copy-pasted verbatim into `CupertinoControls`, `MaterialControls`, and `MaterialDesktopControls`. Not blocking — the three controls already carry a lot of parallel boilerplate — but if a `MixinIdleCursor` (or a top-level `idleCursorFor(chewieController, notifier)` helper) ever gets extracted, this would be a clean candidate.

`hideCursorInFullScreen` defaulted to on-by-default is the right choice: it matches YouTube/VLC/mpv, and users can turn it off via the flag. LGTM.

PR #956, opened 2026-07-11
feat: show a hover-time indicator on the progress bar

Ortes:
### Description

Adds a small floating time indicator to the Material progress bar. While the pointer hovers (or drags) the bar, a pill shows the timecode at that position, so the user can see exactly where a click or seek will jump to — the behaviour people expect from YouTube/Netflix-style scrubbers.

The pill is rendered as an overlay above the bar and follows the pointer horizontally, clamped to the bar's width, with a small caret pointing at the exact spot. It uses the existing `formatDuration` helper, so it matches the `position / duration` label already shown in the controls.

### Behaviour

- Only shows once the video is initialized and its duration is known.
- Driven by hover events, so it never appears on touch devices (which don't emit hover) — mobile behaviour is unchanged. It does also track an in-progress drag, which gives touch users seek feedback while scrubbing.
- Purely additive and self-contained in `progress_bar.dart`; no public API changes.

### Notes

Scoped intentionally to the hover-time readout only — it does not set the hover cursor (that is handled separately in #950).


diegotori:

diegotori (approved):
LGTM

PR #957, opened 2026-07-12
feat: show the controls when play/pause comes from outside the controls

Ortes:
## What

When playback is toggled from **outside** the controls — hardware media keys (▶⏸/F8, handled natively by the browser on web), Android headset/notification controls, the MediaSession API, or the embedding app driving the `VideoPlayerController` — the video starts/stops but `MaterialDesktopControls` gives no visual feedback. Pressing space (which goes through the controls' own `_playPause`) reveals the controls; the media key does not.

With this change any `isPlaying` flip reveals the controls and restarts the hide timer, so the state change is always visible and the controls auto-hide again after `hideControlsTimer`.

## How

`_updateState` (already registered as a listener on the `VideoPlayerController`) compares the incoming `controller.value.isPlaying` against the previous snapshot `_latestValue` — which now gets an initializer so its first read can't throw — and calls the existing `_cancelAndRestartTimer()` on a flip. No new state, no platform-specific code: every platform implementation funnels external play/pause into `value.isPlaying` (`isPlayingStateUpdate`), so this covers web media keys, ExoPlayer's `onIsPlayingChanged`, iOS Control Center, etc.

## Tests

`test/external_playpause_test.dart` drives an externally-mutated controller and covers: reveal + auto-hide on external play, reveal + auto-hide on external pause, and no reveal on position-only updates.

`dart format`, `dart analyze lib test`, and the full `flutter test` suite are clean.

### DIFF
--- a/lib/src/cupertino/cupertino_controls.dart
+++ b/lib/src/cupertino/cupertino_controls.dart
@@ -39,7 +39,7 @@ class CupertinoControls extends StatefulWidget {
 class _CupertinoControlsState extends State<CupertinoControls>
     with SingleTickerProviderStateMixin {
   late PlayerNotifier notifier;
-  late VideoPlayerValue _latestValue;
+  VideoPlayerValue? _latestVideoPlayerValue;
   double? _latestVolume;
   Timer? _hideTimer;
   final marginSize = 5.0;
@@ -56,6 +56,9 @@ class _CupertinoControlsState extends State<CupertinoControls>
   // We know that _chewieController is set in didChangeDependencies
   ChewieController get chewieController => _chewieController!;
   ChewieController? _chewieController;
+
+  // We know that _latestVideoPlayerValue is set in didChangeDependencies
+  VideoPlayerValue get _latestValue => _latestVideoPlayerValue!;
 
   @override
   void initState() {
@@ -145,6 +148,7 @@ class _CupertinoControlsState extends State<CupertinoControls>
     final oldController = _chewieController;
     _chewieController = ChewieController.of(context);
     controller = chewieController.videoPlayerController;
+    _latestVideoPlayerValue = controller.value;
 
     if (oldController != chewieController) {
       _dispose();
@@ -746,8 +750,22 @@ class _CupertinoControlsState extends State<CupertinoControls>
       _displayBufferingIndicator = buffering;
     }
 
+    // Play/pause can also come from outside these controls (Control Center,
+    // the lock screen, headset buttons): reveal the controls the same way
+    // _playPause does, so the state change is visible either way.
+    if (_latestValue.isPlaying != controller.value.isPlaying) {
+      if (controller.value.isPlaying) {
+        _cancelAndRestartTimer();
+      } else {
+        setState(() {
+          notifier.hideStuff = false;
+        });
+        _hideTimer?.cancel();
+      }
+    }
+
     setState(() {
-      _latestValue = controller.value;
+      _latestVideoPlayerValue = controller.value;
       _subtitlesPosition = controller.value.position;
     });
   }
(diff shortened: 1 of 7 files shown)


lib/src/material/material_desktop_controls.dart
```diff
@@ -30,7 +30,7 @@ class MaterialDesktopControls extends StatefulWidget {
 class _MaterialDesktopControlsState extends State<MaterialDesktopControls>
     with SingleTickerProviderStateMixin {
   late PlayerNotifier notifier;
-  late VideoPlayerValue _latestValue;
+  late VideoPlayerValue _latestValue = controller.value;
```
diegotori: This should be moved into either `initState` or `didChangeDependencies` and it should be made nullable.

diegotori (changes requested):

lib/src/cupertino/cupertino_controls.dart
```diff
@@ -56,6 +56,9 @@ class _CupertinoControlsState extends State<CupertinoControls>
   ChewieController get chewieController => _chewieController!;
   ChewieController? _chewieController;
 
+  // We know that _latestVideoPlayerValue is set in didChangeDependencies
+  VideoPlayerValue get _latestValue => _latestVideoPlayerValue!;
```
diegotori: Don't force-unwrap. It should maintain its nullability here too.

diegotori (changes requested):

lib/src/cupertino/cupertino_controls.dart
Ortes: Took a third route: the nullable field and the getter are both gone. `_latestValue` is back to master's `late VideoPlayerValue _latestValue;`, assigned in `didChangeDependencies` one line after `controller` — so the init is where you asked, with no `?`/`!` and no change to the existing reads.

Write-before-read holds because master already writes the field first in `_updateState`, reached from `_initialize()`, which `didChangeDependencies` calls; this only moves that write earlier. `initState` never reads it.

Happy to make it genuinely nullable at every read if you'd rather have that from the type system.

diegotori: That definitely works.

diegotori:
@Ortes please re-sync your changes so that I can merge this PR.

Do likewise for all of your other pending PRs. Thanks.

ibrahim-iqbal (approved):
Solid feedback fix. Detecting the transition by comparing `_latestValue.isPlaying` against the current `controller.value.isPlaying` before updating `_latestValue` catches every source of the play/pause flip — the ones from inside the controls fire the same branch harmlessly, and the ones from outside (media keys, MediaSession, embedding app calling into `VideoPlayerController`) now reveal the controls the same way `_playPause` does.

The pair of branches mirrors the existing `_playPause` semantics correctly:

- On resume (`isPlaying = true`): `_cancelAndRestartTimer()` — shows controls and auto-hides after the standard delay, so the reveal doesn't linger.
- On pause (`isPlaying = false`): reveal + cancel the hide timer, so the controls stay visible until the user does something.

Small notes, non-blocking:

- The added `_latestValue = controller.value;` in `didChangeDependencies` matters — without seeding it the very first `_updateState` frame would compare against a default `VideoPlayerValue` and might see a spurious "flip" that triggers the reveal on initial mount. Good catch.
- The same 15-line block is now duplicated across `MaterialControls`, `MaterialDesktopControls`, and `CupertinoControls`. Same story as #954 — the three controls already carry a lot of parallel state, so a mixin or a shared `_reactToPlaybackToggle` helper would only make sense when a third or fourth signal joins this pattern.
- The added `external_playpause_test.dart` (149 lines) exercises the reveal for both start and stop transitions across all three controls; that's the coverage that will keep future refactors honest.

LGTM.

diegotori (approved):
LGTM.

PR #961, opened 2026-08-19
Casting support (Chromecast, AirPlay or anything else)

a1rwulf:
# PR description

This is an attempt to close the outstanding "Screen-Mirroring / Casting (Google Chromecast)" item on the README
roadmap.

In order to be as lightweight as possible, so not all apps that don't care about casting pull the heavy casting dependencies, I went for the approach to have a generic casting backend that can be done in a separate package.

The work needed is quite extensive, so this is a draft to work towards upstreaming such a feature.
It has several prerequisites described in detail below, but the summary is:
* Airplay needs an upstream patch PRed here: https://github.com/flutter/packages/pull/12490
* Casting in general needs the companion package here: https://github.com/a1rwulf/chewie_cast/tree/main
* The companion package needs the latest release of: https://github.com/felnanuke2/flutter_google_cast as it contains needed fixes found during development

Detailed description of the approach follows:

Chewie gains the casting *UI* and the local↔remote handover, but no sender. The
backend is an abstract class the app implements, so a pure-Flutter video player
package does not drag the Google Cast SDK — an iOS static framework, an Android
Play Services dependency and a per-platform setup — into every app that just
wants play controls. **No new runtime dependencies**; the only pubspec change is
a dev dependency on `video_player_platform_interface` for the fake platform the
widget tests install.

## What an app implements

`ChewieCastController` is a `ChangeNotifier` covering discovery (`devices`,
`startDiscovery`), session (`connect`, `disconnect`, `connectionState`) and
remote playback (`load`, `play`, `pause`, `seekTo`, `setVolume`, expressed as a
`VideoPlayerValue` so the existing controls can read it unchanged).

```dart
ChewieController(
  videoPlayerController: controller,
  castController: myCastController,
  castMedia: CastMedia(url: '…', mimeType: 'video/mp4', title: 'Big Buck Bunny'),
);
```

## What Chewie does with it

* A cast button in all three skins, which pulses while connecting and switches
  glyph once connected.
* A device picker, and an overlay replacing the video while the receiver has it.
* **Handover.** On connect the local player pauses and the receiver is loaded at
  the local position; on disconnect the local player resumes at the position the
  receiver reached. Transport controls and the progress bar retarget to whichever
  is playing, via a small `ChewiePlaybackTarget` indirection.
* A session that is *already* running when a `ChewieController` is built is
  adopted, rather than noticed only if it changes afterwards. Senders belong to
  the app and outlive the screens that create them, so without this, opening a
  second video while casting plays it on the device while the receiver still
  holds the first.
* Controls stay revealed while casting rather than fading, since there is no
  video underneath to look at.

All of it is inert when `castController` is null — the default — so nothing
changes for existing users.

## Also here

`additionalControls` puts app-supplied widgets in the control bar, for controls
that have to be a widget rather than an options-sheet row — an AirPlay button
being the motivating case. `ChewieControlStyle` lets a bar tell those widgets
how it dresses its own buttons — icon size, tint, padding, chrome — so one
widget matches all three skins instead of being sized for one and sitting wrong
in the others. The example demonstrates both with a dependency-free control.

## Verification

114 widget tests, covering handover in both directions, retargeting, the picker,
session adoption, and the styling contract.

Beyond the tests, this was driven against real hardware — an NVIDIA Shield from
iOS via a companion package implementing this API against the Cast SDK, plus
AirPlay: discovery, session, handover at position, playback, remote status
mirrored back into the controls, restart after the receiver finishes, and
disconnect.

That found defects the tests could not, including a progress bar that asserts
when a source reports itself initialized before it knows its duration — which
cast receivers do routinely, and which is fixed here — and the already-running
session case above, which only shows up once a sender outlives a screen.

## Companion package

`chewie_cast` implements this API for Google Cast and AirPlay, and stays out of
this repo deliberately. Happy to adjust the naming or the shape of the
abstraction — the split is the part worth agreeing on.
You can find the companion package source here: https://github.com/a1rwulf/chewie_cast/tree/main

### Without any hardware or extra packages

`flutter run` in `example/`. It ships `DemoCastController`, a fake backend that
simulates discovery, a connection delay and a receiver whose position ticks on
its own. That exercises the whole surface — cast button states, the picker,
handover in both directions, the overlay, the controls retargeting to the
receiver — with nothing to install. Switching skins in the example shows the
same control bar in Material, Material desktop and Cupertino.

### With a real Chromecast

The sender lives in `chewie_cast`. 

```yaml
dependencies:
  chewie_cast:
    git:
      url: https://github.com/a1rwulf/chewie_cast.git
      ref: main

dependency_overrides:
  chewie:
    path: ../
```

Then swap the fake for the real one in `example/lib/app/app.dart` — it is a
one-line change, since both implement `ChewieCastController`:

```dart
final ChromecastController _castController = ChromecastController();
// …and await _castController.initialize() before building the ChewieController.
```

The Cast SDK needs per-platform setup (Bonjour entries on iOS, an options
provider on Android) or discovery silently finds nothing; `chewie_cast`'s README
has the exact blocks. Note that the receiver fetches the media URL itself, so it
must be reachable from the TV and served with CORS headers — Google's Default
Media Receiver requires them, and without them playback fails with an `idle`
status that names no reason.

### AirPlay on iOS

AirPlay additionally needs a patched `video_player_avfoundation`. Stock
`video_player` never sets `allowsExternalPlayback`, so selecting an Apple TV
mirrors the phone screen rather than handing the video to it. That is two lines
and an open PR against flutter/packages; until it lands:

```yaml
dependency_overrides:
  video_player_avfoundation:
    git:
      url: https://github.com/a1rwulf/flutter_packages.git
      path: packages/video_player/video_player_avfoundation
      ref: seadev_airplay-external-playback
```

The mentioned upstream PR: https://github.com/flutter/packages/pull/12490

Fixes:
https://github.com/fluttercommunity/chewie/issues/511
https://github.com/fluttercommunity/chewie/issues/329


a1rwulf:
Would be nice to get the ball rolling on this one, most current video players support casting or airplay - this is one of the missing key features in chewie for quite some time, so I would like to get this over the finish line.

Especially interested in what you think about the generic backend approach to keep casting dependencies out of the main package.

diegotori:
@a1rwulf please fix your conflicts. Thanks.

diegotori:
@a1rwulf thanks for your contributions thus far.

I'm assuming that you'll publish your companion package once it's ready on your end?

lib/src/cupertino/cupertino_controls.dart
```diff
@@ -537,6 +556,47 @@ class _CupertinoControlsState extends State<CupertinoControls>
     );
   }
 
+  Widget _buildCastButton(
```
diegotori: Instead of a method that returns a widget, create a standalone widget that encapsulates this implementation. Then add it to wherever you called this method.

lib/src/cupertino/cupertino_controls.dart
```diff
@@ -560,6 +620,49 @@ class _CupertinoControlsState extends State<CupertinoControls>
...
+                    // Each supplied control wears its own frosted pill, so
+                    // they need the gap the bar's own buttons have between
+                    // them; without it two of them sit pill against pill.
+                    for (final control in chewieController.additionalControls!(
+                      context,
+                    )) ...[control, SizedBox(width: marginSize)],
+                  ],
+                ),
+              ),
+            ),
```
diegotori: Same here.

lib/src/cupertino/cupertino_controls.dart
```diff
@@ -699,15 +811,19 @@ class _CupertinoControlsState extends State<CupertinoControls>
...
-      controller.setPlaybackSpeed(selectedSpeed);
+      _playback.setPlaybackSpeed(selectedSpeed);
     });
   }
 
   void _startHideTimer() {
+    // While casting the phone is a remote control, not a video surface: there
+    // is nothing behind the controls worth revealing, and hiding them costs the
+    // user a tap, because the first one only brings them back.
+    if (chewieController.isCasting) return;
```
diegotori: I'm wondering if it makes sense to display something over the video overlay while casting. That might be a use case for some users of your code.

Consider exposing an optional builder from `chewieController` that if present, displays an overlay over the video while casting.

lib/src/material/material_controls.dart
```diff
@@ -144,6 +162,29 @@ class _MaterialControlsState extends State<MaterialControls>
...
+                  iconColor: Colors.white,
+                  padding: const EdgeInsets.all(12),
+                  decorate: (child) => child,
+                  child: Row(
+                    mainAxisSize: MainAxisSize.min,
+                    children: chewieController.additionalControls!(context),
+                  ),
+                ),
+              if (_castController != null && chewieController.allowCasting)
+                CastButton(
```
diegotori, line 195: Consider creating a shared `CastButton` that can be configured differently based on whether you're using Material or Cupertino. That way, it falls under the same component.

That and you can create a platform specific version for both UIs.

lib/src/material/material_controls.dart
```diff
@@ -583,6 +634,10 @@ class _MaterialControlsState extends State<MaterialControls>
   }
 
   void _startHideTimer() {
+    // While casting the phone is a remote control, not a video surface: there
+    // is nothing behind the controls worth revealing, and hiding them costs the
+    // user a tap, because the first one only brings them back.
+    if (chewieController.isCasting) return;
     final hideControlsTimer = chewieController.hideControlsTimer.isNegative
```
diegotori, line 670: See the above note regarding exposing a widget builder function that if present, displays in the built widget above the video, some UI when casting.

a1rwulf:
@diegotori yes of course.
Since https://github.com/flutter/packages/pull/12490 was accepted upstream, my companion package will enable airplay support once this PR is merged.

Question is, if you are OK with:
a) the general approach of using a companion package
b) the code quality meets your expectations

Once those questions are settled we can think of how to organize merging of this PR.
I wouldn't mind publishing my package prior to merging this PR, so once it hits chewie it actually does something.

Secondly - without knowing the rules about the fluttercomunity org - I can also donate the companion package, so it lives under the same umbrella as chewie itself.

Let me know your thoughts.

diegotori (changes requested):

diegotori:
> Question is, if you are OK with: a) the general approach of using a companion package b) the code quality meets your expectations

It's getting there. See the comments thus far. However, since you added test coverage for this component, then it's on the way to looking good.

> Once those questions are settled we can think of how to organize merging of this PR. I wouldn't mind publishing my package prior to merging this PR, so once it hits chewie it actually does something.

Once I release these changes in a future release (probably `1.17.0`), then that should suffice for releasing that library.

> Secondly - without knowing the rules about the fluttercomunity org - I can also donate the companion package, so it lives under the same umbrella as chewie itself.

Please read the existing [transfer guide](https://github.com/fluttercommunity/transfer-guide) for steps on how to do this. You may have to also reach out to @slightfoot for this as well, since he's one of the primary managers of this org.

lib/src/material/material_controls.dart
a1rwulf: A cast overlay is already implemented here:
cupertino:826 / material:641 — `castOverlayBuilder` already exists on `ChewieController`, used in player_with_controls.dart with a `CastOverlay` default.

lib/src/cupertino/cupertino_controls.dart
a1rwulf: See https://github.com/fluttercommunity/chewie/pull/961#discussion_r3908005787
I did implement a cast overlay.

lib/src/cupertino/cupertino_controls.dart
a1rwulf: Done.

lib/src/material/material_controls.dart
a1rwulf: Material and Material desktop already use the shared CastButton — that's the CastButton on the line you commented on. Cupertino now has CupertinoCastButton, which wraps the same component with the frosted pill and the fading Cupertino's bar needs, so both skins go through one component.

I did not add a MaterialCastButton, since Material adds no chrome and it would be an empty pass-through. Happy to add it if you want the symmetry.

a1rwulf:
While working on your review items, I noticed my cast devices picker was material only.
I have now added a cupertino style cast device picker.
I'll do another runtime test tomorrow and give green light once I'm done testing with real equipment.

a1rwulf:
> Once I release these changes in a future release (probably 1.17.0), then that should suffice for releasing that library.

I actually meant it the other way round, I should publish the companion package first - even though it has no use on it's own, but it will immediately enable users to use casting in 1.17.0.
If you release before I publish the companion package, you have casting in the release notes but it's a noop until my package is published.

But now that I know you're interested in getting this merged, I can finalize the package as well - it's verified to work, but probably needs some polishing here and there.

diegotori:
@a1rwulf once I publish, I'll link to the github repo. From there you can freely publish at your earliest convenience.

If there's still polish to be done, then I would suggest placing this PR in Draft until you're ready for another review.

a1rwulf:
OK, I e2e tested the latest version of the PR successfully.
Would be nice to finalize the review.

In order to get this working E2E for chewie consumers, https://github.com/flutter/packages/pull/12490 needs to land in a vp release.
So I will put the PR to draft until that happened and publish the companion package meanwhile.
Once everything is ready, I will bump the dependency on vp and mark the PR as ready.

Does that sound like a plan?

Edit:
My plan has a weakness, my packageg needs to depend on a chewie that has this change already.
So my package is end of the chain.

diegotori:
@a1rwulf converting your PR to Draft until that `video_player` PR gets deployed into a new release. Feel free to mark it Ready for Review once that's done.

Then once this goes out in a future release, I'll pull in that new version and you'll be able to release yours afterwards.

a1rwulf:
> @a1rwulf converting your PR to Draft until that `video_player` PR gets deployed into a new release. Feel free to mark it Ready for Review once that's done.
> 
> Then once this goes out in a future release, I'll pull in that new version and you'll be able to release yours afterwards.

Perfect.
Thank you, e2e testing casting found a bug in the chewie sample, I just included a fix for it in this branch - fyi.
Once I'm satisfied with the e2e testing, I will mark this PR ready for review.
Publishing of my package will also be ready when VP is released.

Thanks for your help!

a1rwulf:
Just as a fyi:
I noticed that during a cast session, chewie was drawing the usual loading spinner on top of the casting overlay, which looks a bit awkward, hence the new commit which gives chewie a way to know that playback is routed to a remote and omit drawing the spinner over the casting overlay.
I need to do another e2e run with Chromecast equipment, will set the branch to ready for review once that's done.

a1rwulf:
Ok the upstream PR was merged and meanwhile I have my package ready.
The 3 new commits were necessary because of some cosmetics I wanted to fix properly:

1. Casting takes some time and the OSD vanished so the user did not notice what's going on
2. Loading and play/pause overlay drawn above the casting overlay which looks weird, so it's hidden. Users can see that casting is loading on the TV they are streaming too. No need to render a loading animation over the casting overlay

a1rwulf:
video_player_av_foundation 2.12.0 is already tagged as well.
So the prerequisite for this PR is green.

diegotori:
@a1rwulf please fix re-sync your changelog with the latest upstream changes. Thanks.

a1rwulf:
> @a1rwulf please fix re-sync your changelog with the latest upstream changes. Thanks.

Done.

diegotori (approved):
LGTM.

a1rwulf:
Awesome, thanks @diegotori - here is the promised backend package: https://pub.dev/packages/chewie_cast
Which means, everything is fully functional without any downstream workarounds.
Casting is officially off the todolist.

PR #779, opened 2023-10-03
Fix(example): Toggle video source many times throws PlatformException on Android

magno-castro:
I runned the package last version (1.7.1) example in my device  (Poco F1) and if I press the "toggle video src" button about 8 times, it throws the PlatformException above:
Screenshot 2023-10-03 at 14 32 42

I noticed that `toggleVideo` function calls `initializePlayer`, but without dispose controllers.

So, dispose controllers before reinitialize fix this problem.

```
[✓] Flutter (Channel stable, 3.13.6, on macOS 13.4.1 22F770820d darwin-arm64,
    locale en-BR)
    • Flutter version 3.13.6 on channel stable at
      /Users/magnocastromoraes/Development/sdks/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision ead455963c (7 days ago), 2023-09-26 18:28:17 -0700
    • Engine revision a794cf2681
    • Dart version 3.1.3
    • DevTools version 2.25.0

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
    • Android SDK at /Users/magnocastromoraes/Development/sdks/android
    • Platform android-33, build-tools 33.0.1
    • ANDROID_HOME = /Users/magnocastromoraes/Development/sdks/android
    • ANDROID_SDK_ROOT = /Users/magnocastromoraes/Development/sdks/android
    • Java binary at: /Applications/Android
      Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build
      11.0.13+0-b1751.21-8125866)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14E222b
    • CocoaPods version 1.12.0

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2021.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build
      11.0.13+0-b1751.21-8125866)

[✓] VS Code (version 1.82.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
[...]

PR #963, opened 2026-08-28
Add ChewieController.swapVideoSource (in-place video source replacement)

Ortes:
## What this adds

`ChewieController.swapVideoSource(VideoPlayerController newController, {bool disposeOldController = true})` — replaces the internal `VideoPlayerController` in place while preserving **position, playback speed, volume, looping and play/pause state**. The new controller is initialized *before* the swap (the player never shows an unready source; on failure the error is rethrown and the old controller stays), and the old controller is disposed *after* the UI has re-attached. `disposeOldController: false` opts out, for hosts that pool preloaded controllers.

To make an in-place swap actually reach the screen, `PlayerWithControls` now listens to the `ChewieController` (the `InheritedWidget` only notifies on controller *identity* changes) and keys the controls subtree on the video controller, so the control skins are recreated and re-attach their listeners to the new controller. `VideoPlayer` itself re-attaches through its own `didUpdateWidget`.

## Bugs fixed on the way

- **#618** — `ChewieState.listener()` popped the current route on *any* `notifyListeners()` while fullscreen (`else if (_isFullScreen)`), not just on an actual exit-fullscreen request. Since `swapVideoSource` notifies, this fix is a prerequisite; it also fixes the original speed-dialog report.
- `Chewie.didUpdateWidget` re-added its listener to the new controller without removing it from the old one (listener leak).

## Why

This is groundwork for quality switching (see the follow-up PR). #492 and #867 both needed to replace the controller in place and both struggled with the same three problems this PR isolates and tests: preserving playback state, getting the UI to re-attach, and disposing the old controller safely — the missing dispose is the crash documented in #779. Thanks @merdweebner (#492) and @akmalova (#867) for the groundwork.

## Tests

New `FakeVideoPlayerPlatform` (dev-dep on `video_player_platform_interface` only) so controllers can actually initialize under `flutter test`, plus 7 tests covering: state preservation, single notification, deferred disposal of the old controller, the `disposeOldController: false` opt-out, initialization-failure rollback, video/controls re-attachment, and a #618 regression test (swap while fullscreen keeps the route).

### DIFF
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -78,6 +78,7 @@ class ChewieState extends State<Chewie> {
   @override
   void didUpdateWidget(Chewie oldWidget) {
     if (oldWidget.controller != widget.controller) {
+      oldWidget.controller.removeListener(listener);
       widget.controller.addListener(listener);
     }
     super.didUpdateWidget(oldWidget);
@@ -93,7 +94,7 @@ class ChewieState extends State<Chewie> {
       _resumeAppliedInFullScreen = false;
       _isFullScreen = isControllerFullScreen;
       await _pushFullScreenWidget(context);
-    } else if (_isFullScreen) {
+    } else if (!isControllerFullScreen && _isFullScreen) {
       Navigator.of(
         context,
         rootNavigator: widget.controller.useRootNavigator,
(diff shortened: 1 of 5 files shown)


diegotori (approved):
LGTM.

PR #966, opened 2026-08-28
Fix progress bar handle snapping back during a seek

Ortes:
## Problem

Dragging the progress bar handle and releasing it makes the handle jump back to the previous playback position for a few frames before landing on the target. The same lag affects a plain tap on the bar, where the handle does not move until the seek finishes.

`VideoPlayerController.seekTo()` only writes the new position into `VideoPlayerValue.position` *after* the platform reports the seek complete:

```dart
await _videoPlayerPlatform.seekTo(_playerId, position);
_updatePosition(position);
```

`_VideoProgressBarState` cleared `_latestDraggableOffset` synchronously right after firing the seek, so for the whole duration of the platform round-trip the painter fell back to the stale `value.position`.

The window is longest on iOS — `FVPVideoPlayer.seekTo` calls `AVPlayer.seekToTime` with `kCMTimeZero` tolerance, so it waits for the exact target frame to be decoded — which is where it was reported, but the flicker exists on every platform.

A second, smaller issue: `controller.play()` was called *before* the seek was issued, so the player briefly resumed at the old position (visible in both the video and the audio) while the seek was still in flight.

## Fix

- Track the requested position in `_pendingSeekPosition` and keep painting it until `seekTo()` resolves. Paint priority is now active drag offset → pending seek position → reported position. Both of the first two derive the same `Duration`, so the handoff on finger-release is pixel-identical and there is no visible transition.
- A `_latestSeekRequestId` counter so a slow seek resolving after a newer one cannot clear the newer position.
- Taps go through the same path, so the handle moves to the tapped point immediately.
- Resume playback only once the seek has landed. `onDragEnd` still fires synchronously first, so hide-timer behaviour in the controls is unchanged.

`StaticProgressBar` gains an optional `pendingSeekPosition` parameter; the existing `latestDraggableOffset` is untouched, so this is not a breaking change.

## Tests

Four widget tests in `test/progress_bar_seek_test.dart`, driven by a fake `VideoPlayerPlatform` whose seeks complete only when the test says so — which is what makes the platform latency observable:

- the dropped position stays painted while the controller still reports the old one
- playback resumes only after the seek lands
- a tap paints the tapped position immediately
- a stale seek does not clear the position requested by a newer one

This adds `video_player_platform_interface` as a dev dependency, needed to implement the fake.

`flutter test` passes, `flutter analyze` is clean, and the code is `dart format`ted.

## Verification

Confirmed fixed on a physical iPhone (iOS 26.5, release build) with the example app, on both the Cupertino and Material controls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [Unreleased]
+* 🛠️ Fix the progress bar handle snapping back to the previous position for the duration of a seek (most visible on iOS, where an exact seek is slow to complete). The requested position is now painted until the controller reports it, and playback resumes only once the seek has landed.
+
 ## [1.15.0]
 * 🌐 [#946](https://github.com/fluttercommunity/chewie/pull/946): Web: enter the browser's native (OS-level) fullscreen via the Fullscreen API instead of only expanding the Flutter view inside the browser window. Pressing Escape to leave browser fullscreen also exits Chewie's fullscreen. Controlled by the new `ChewieController.useNativeFullScreenOnWeb` flag (defaults to `true`; no effect on non-web platforms). Thanks [Ortes](https://github.com/Ortes).
 * 🖱️ [#950](https://github.com/fluttercommunity/chewie/pull/950): Show click cursor on hover over Material controls and progress bar. Thanks [Ortes](https://github.com/Ortes).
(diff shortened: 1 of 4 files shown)


diegotori (approved):
LGTM.

PR #970, opened 2026-09-07
fix(web): fullscreen pops the host route on iPhone Safari (no Fullscreen API)

### LINKED ISSUE #969: Web: exiting fullscreen pops the host route on iPhone Safari (requestFullscreen missing + blind Navigator.pop)
## Summary

On iPhone Safari (Flutter web), toggling fullscreen twice pops the **embedder's** current route instead of Chewie's fullscreen route. In our app this exits the page hosting the player and loses the user's in-progress state.

Two independent defects combine:

1. **`requestBrowserFullscreen()` throws on iPhone Safari.** Since #946 (`useNativeFullScreenOnWeb`, default `true`, shipped in 1.15.0), `_pushFullScreenWidget` calls `web.document.documentElement?.requestFullscreen()`. `Element.requestFullscreen` does not exist on iPhone Safari (only `HTMLVideoElement.webkitEnterFullscreen` does; iPad is fine), so this raises a JS `TypeError`. It happens **after** `ChewieState._isFullScreen` and the controller flag were set to `true` but **before** the fullscreen route is pushed. State and navigator are now out of sync.
2. **The exit path pops blindly.** `ChewieState.listener()` runs `Navigator.of(context, rootNavigator: useRootNavigator).pop()` without checking that the route on top is the one Chewie pushed. Combined with the desync above, the second tap pops the embedder's page. This is the same mechanism as #618.

## Reproduction

- Flutter web build, `ChewieController` with defaults (`useNativeFullScreenOnWeb: true`, `useRootNavigator: true`).
- Open it in Safari on an iPhone, tap the fullscreen button: nothing visible happens, the icon flips to "exit fullscreen", and the console shows `TypeError: ...requestFullscreen is not a function`.
- Tap the fullscreen button again: the page hosting the player is popped.

## Proposed fix

- Guard the browser request: only call `requestFullscreen` when the element exposes it (`document.fullscreenEnabled` / `'requestFullscreen' in element`); otherwise skip native browser fullscreen and keep the Flutter-only fullscreen route.
- Never pop a route Chewie did not push: keep the `PageRouteBuilder` instance created in `_pushFullScreenWidget` and, on exit, call `Navigator.removeRoute(route)` (or check `route.isCurrent` before `pop()`).
- Optional follow-up: on mobile browsers, offer the `<video>` element's own fullscreen (`video.requestFullscreen()` on Android Chrome, `video.webkitEnterFullscreen()` on iPhone Safari) behind a flag, building on #947 (`disableFullScreenRoute`). We currently do this on the app side.

Happy to open a PR for the first two points.

Ortes:
Fixes #969.

> **Urgent, please.** This is a user-facing regression introduced by the `useNativeFullScreenOnWeb` default in 1.15.0, not a new feature. Every app shipping Chewie on Flutter web is affected on iPhone Safari: tapping the fullscreen button navigates the user off the page holding the player. In our case patients lost an in-progress exercise session mid-treatment, and we are carrying an app-side workaround until this lands. A patch release would let us and other web users drop it.

### The bug

Since 1.15.0 `useNativeFullScreenOnWeb` defaults to `true`, and `_pushFullScreenWidget` calls `document.documentElement.requestFullscreen()` before pushing the fullscreen route. iPhone Safari has no `Element.requestFullscreen` (only `HTMLVideoElement.webkitEnterFullscreen`), so the call throws. By then `ChewieState._isFullScreen` and the controller flag are already `true`, and no route was pushed. The next tap flips the controller flag back, the listener sees "widget says fullscreen, controller says not", and runs `Navigator.pop()` on the root navigator. The route on top is the embedder's own page, which is what gets popped.

From the user's seat: first tap does nothing, second tap throws them out of the page. Same mechanism as #618.

Browsers that have the API are unaffected: the route is pushed, and the exit pops that route. Android Chrome and desktop are fine, verified in production.

### The fix

On a browser without the Fullscreen API, Chewie does not push or pop anything. It opens the video element's own player instead:

- `browserFullscreenSupported`: feature check on `documentElement`.
- `listener()`: when the check fails (and `useNativeFullScreenOnWeb` is on), mirror the controller flag, call `webkitEnterFullscreen()` / `webkitExitFullscreen()` on the element `video_player_web` created for the controller's `playerId`, and return.
- The element's `webkitendfullscreen` event calls `exitFullScreen()`, so the control icon follows the user closing the native player.

The route path, `_reInitializeControllers`, and every platform that has the API are untouched. Users on iPhone also get a better result than the Flutter route would have given them: the phone's own fullscreen player, with no `<video>` reparenting and so no stream reload on enter or exit.

### Scope

Deliberately minimal, +85 lines, nothing removed. Only the code path that threw is diverted, and it is gated on `useNativeFullScreenOnWeb`, so a host that opted out of browser fullscreen keeps the Flutter route on iPhone exactly as before.

### Tests

`test/video_element_fullscreen_test.dart`, driven by a new `@visibleForTesting` `ChewieState.debugUsesVideoElementFullScreen` override. `kIsWeb` is a compile-time constant, so this branch is otherwise unreachable from a VM test; the override follows the spirit of `debugDefaultTargetPlatformOverride`. A `NavigatorObserver` counts pushes and pops:

- entering fullscreen pushes no route;
- leaving fullscreen pops nothing, so the host page stays mounted;
- a redundant exit request also pops nothing, which is the failure users hit;
- and the opposite case is pinned too: where the browser has the API, the route is still pushed and popped.

Patch coverage goes from 11% to roughly 82%. What is left uncovered is the `kIsWeb` conjunction itself and the stub getter it short-circuits, which no VM test can reach. The web helper (`web_fullscreen_impl.dart`) is not compiled off web, so it is not part of the report.

`flutter analyze` clean, full suite green (104 tests), `flutter build web` on the example passes.

### One thing I would like your call on

Locating the `<video>` element needs `VideoPlayerController.playerId`, which `video_player` marks `@visibleForTesting` with a note that plugin consumers should not use it, hence the `ignore` on that getter. I could not find a public way to map a controller to its element, and this is what apps are doing today in userland. If you would rather see a different approach, or want to raise it with `video_player` for a public accessor, say so and I will follow.

### Device check

Not run on a physical iPhone from here. The element calls mirror what our app has been shipping for the same `<video>` (`webkitEnterFullscreen`, and closing the native player returns inline with the position kept), so a confirmation from a reviewer with an iPhone would be welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [Unreleased]
+* 🐛 [#970](https://github.com/fluttercommunity/chewie/pull/970): Web: fix fullscreen popping the host route on browsers without the Fullscreen API. iPhone Safari has no `Element.requestFullscreen`, so entering fullscreen threw before the route was pushed and the next exit request popped the embedder's page. Those browsers now use the video element's own fullscreen player and no route is involved. Thanks [Ortes](https://github.com/Ortes).
+
 ## [1.16.2]
 * ⬆️ [#959](https://github.com/fluttercommunity/chewie/pull/959): Add YouTube-style chapters to the progress bar. Pass `ChewieController.chapters` to split the bar into chapter segments and show the hovered/scrubbed chapter title above it. Thanks [Ortes](https://github.com/Ortes).
 * 🛠️ [#968](https://github.com/fluttercommunity/chewie/pull/968): Reformat `_PlaybackSpeedDialog` to satisfy `dart format`. Thanks [Ortes](https://github.com/Ortes).
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -49,14 +49,23 @@ class ChewieState extends State<Chewie> {
   late PlayerNotifier notifier;
   late final void Function() _browserFsExitHandler;
 
+  /// Forces the video-element branch on or off, for tests.
+  ///
+  /// [kIsWeb] is a compile-time constant, so the branch below is unreachable
+  /// from a VM test without an override, in the spirit of
+  /// `debugDefaultTargetPlatformOverride`. Reset it to null when done.
+  @visibleForTesting
+  static bool? debugUsesVideoElementFullScreen;
+
   /// Browsers without the Fullscreen API (iPhone Safari) cannot host the
   /// fullscreen route: `requestFullscreen` throws before the route is pushed,
   /// and the next exit request pops the page underneath instead. On those
   /// browsers the video element's own player is used and no route is involved.
   bool get _usesVideoElementFullScreen =>
-      kIsWeb &&
-      widget.controller.useNativeFullScreenOnWeb &&
-      !browserFullscreenSupported;
+      debugUsesVideoElementFullScreen ??
+      (kIsWeb &&
+          widget.controller.useNativeFullScreenOnWeb &&
+          !browserFullscreenSupported);
 
   // ignore: invalid_use_of_visible_for_testing_member
   int get _playerId => widget.controller.videoPlayerController.playerId;
--- a/lib/src/chewie_player.dart
+++ b/lib/src/chewie_player.dart
@@ -49,6 +49,18 @@ class ChewieState extends State<Chewie> {
   late PlayerNotifier notifier;
   late final void Function() _browserFsExitHandler;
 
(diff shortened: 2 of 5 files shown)


diegotori (approved):
LGTM.

